PR #10, opened 2015-12-06
Add appDisplayName static var.

domhof:


goktugyil:
Awesome! Thanks :+1: 
I'll include in this in v1.0

PR #11, opened 2015-12-19
Extension for UIViewController - to add view controller as child view…

sagarmusale8:
Added extension to add view controller's view as child view controller to any subview of current view controller.


goktugyil:
Awesome, thanks!

PR #14, opened 2015-12-23
Add shake animation to the view extension

avismarahl:
You can use this method to shake a view -- say a textfield if its validation fails.


goktugyil:
Nice man thanks :+1:

PR #16, opened 2015-12-24
Fixed DoubleExtension not compiling with Carthage

fbuentello:
DoublesExtension was not part of the Compile Sources therefore not being built with Carthage.

 - Added DoubleExtensions.swift to Compile Sources in Build Phases    
 - Updated version to 1.0.2


goktugyil:
Thanks man :+1:

PR #24, opened 2016-01-06
Added iOS9 Support for requestURL

indevizible:
//TODO: Seems this is depreceated, update it


goktugyil:
Thanks for the request :+1:

PR #29, opened 2016-01-17
Add default parameter values for Gesture's extension

furuyan:
Gesture's method more simply


goktugyil:
Thanks for the pull request :+1:

PR #43, opened 2016-02-10
Added toBool method in String

lfarah:


goktugyil:
whats the use case here? maybe it should return false when the string is not true? maybe it should return nil? depends on the use case

Issue #19, opened 2015-12-28
Document the code


avismarahl:
Most of the code seems un-documented. We should add an appledoc style documentation.

goktugyil:
Yeah, its definitely needed. There is a good extension which helps with documentation. I can start doing it from smaller files and people might contribute as they find time?

avismarahl:
VVDocumenter, isn't it? Yup, will contribute as I get tim.

goktugyil:
Yeah thats its.

goktugyil:
It seems VVDocumenter isn't needed, shorter documentation is better. I did the first one:

"EZSE:" is just there to clearly identify from UIKit

https://github.com/goktugyil/EZSwiftExtensions/commit/c8ed6e3a6fbd5e18d494767a49b25fce922b2ae6
https://github.com/goktugyil/EZSwiftExtensions/commit/9dfdbd8f4db4426f09254d8782f04400e29be2a4

lfarah:
Im working on that too!

PR #52, opened 2016-02-16
Updated degreesToRadians documentation

lfarah:


lfarah:
#19

Sources/CGFloatExtensions.swift
```diff
@@ -13,7 +13,7 @@ extension CGFloat {
     public var center: CGFloat { return (self / 2) }
 }
 
-/// EZSwiftExtensions
+/// EZSE: Converts angle degrees to radians
```
thellimist: I believe the correct grammar would be `Converts degrees to radians`.

lfarah: @thellimist I think "angle degree's" would be better because anything can be a degree, like temperature

PR #67, opened 2016-02-23
Added makeCircular() to UIViewExtensions

lfarah:


goktugyil:
There was a func which does this

lfarah:
@goktugyil there's an extension for UIImageView called roundSquareImage(), but in UIView would be much better, so not only images can be rounded

goktugyil:
why do you need to set the center?

lfarah:
You don't, sorry. I found that method in the internet and someone added the center

lfarah:
I changed

PR #76, opened 2016-02-25
Int autoConvert()

kevindpsilitonga:
#### Automatically convert your `Int` to whatever type is initializable from `Int` with just one function! 
For example:
```
let myInt = 42
let myDouble: Double = myInt.autoConvert()
let myCGFloat: CGFloat = myInt.autoConvert()
let myFloat: Float = myInt.autoConvert()
let myBool: Bool = myInt.autoConvert()
```
- Saves you time from having to figure out which `.toSomeType` you have to choose.
- Saves you time from having to migrate the code when some type constraint related to `Int` conversion changes in a way that makes your code incompatible

Note: This is meant to be used side-by-side with `.toSomeType` functions since `.toSomeType` is really powerful in circumstances where the type cannot be inferred by `.autoConvert()`

Btw, If you think it's going to result in longer code because you'd have to declare the type, please read the "Longer Explanation" section to see the rationale.

## Longer explanation
Let's say you have this awesome function from an awesome library:
```
func doAwesomeStuff(number: CGFloat)
```

And since you're using integer all over the app, what you do this `doAwesomeStuff(myInt.toCGFloat)` everywhere

However, one day, the awesome library owner decided to change the function definition to this:
```
func doAwesomeStuff(number: Double)
```
Now you have to go to all `doAwesomeStuff` calls you did that involves an `Int` and change
`doAwesomeStuff(myInt.toCGFloat)` to `doAwesomeStuff(myInt.toDouble)`. While it might not be that time consuming for smaller apps, this could be a reason for larger, math heavy apps to not migrate to the newer versions of that awesome library.

#### Good news
`doAwesomeStuff(myInt.autoConvert())` will be suitable for both the version of the library. `.autoConvert()` lets the compiler (as opposed to you) choose which `.toSomeType` function is best so you don't have to change a single line of code during this migration.

### Pros
- Don't have to think about which `.toSomeType` function to choose.
- More robust to changes related to type conversion.

### Cons
- When the type is not defined yet, it'll result in a longer code. For example: `let myDouble = myInt.toDouble` vs.  `let myDouble: Double = myInt.autoConvert()`

### Ideas for Future Improvements
- Add more auto-convertible types. (e.g. Add `.autoConvert()` for `Double`)
- Auto convert functions! (e.g. `let integerDoAwesomeStuff = autoConvert(floatDoAwesomeStuff)`

Feel free to add more suggestions, pros and cons, etc!

### DIFF
--- a/EZSwiftExtensions/IntAutoConvertExtension.swift
+++ b/EZSwiftExtensions/IntAutoConvertExtension.swift
@@ -1,0 +1,33 @@
+import UIKit
+
+protocol InitializableFromInt {
+    init(_ value: Int)
+}
+
+extension Int {
+    /// Automatically converts an `Int` to the right type.
+    func autoConvert<T: InitializableFromInt>() -> T {
+        return T(self)
+    }
+}
+
+// MARK: - FloatingPointType
+extension Double: InitializableFromInt { }
+
+extension Float: InitializableFromInt { }
+
+extension CGFloat: InitializableFromInt { }
+
+// MARK: - SignedIntegerType
+extension Int8: InitializableFromInt { }
+
+extension Int32: InitializableFromInt { }
+
+extension Int64: InitializableFromInt { }
+
+// MARK: - BooleanType
+extension Bool: InitializableFromInt {
+    init(_ value: Int) {
+        self = value != 0
+    }
+}
--- a/EZSwiftExtensionsExample.xcodeproj/project.pbxproj
+++ b/EZSwiftExtensionsExample.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		C85840ED1C43B05200595696 /* NSURLExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = C85840EC1C43B05200595696 /* NSURLExtensions.swift */; };
+		DEF463C31C7ECB6000EC0856 /* IntAutoConvertExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = DEF463C21C7ECB6000EC0856 /* IntAutoConvertExtension.swift */; };
 		E1839DB91BF79335002212C6 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1839DB81BF79335002212C6 /* AppDelegate.swift */; };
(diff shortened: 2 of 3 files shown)


goktugyil:
1. This looks cool and useful, but how about type to type conversions? Double to Int? String to Int? etc.. Would we have to add a similar functionality to every types? 

2. Can we create a general master "Number" type? So autoConvert() changes the thing, what ever it is to the general number and then changes back to whatever needed automatically? 

3. There are merge conflicts now :disappointed:

lfarah:
[Ray Wenderlich's Swift Style Guide](https://github.com/raywenderlich/swift-style-guide#type-inference) talks about type declarations and how we shoudn't do this:
``` swift
let message: String = "Click the button"
```
but instead, do this:
``` swift
let message = "Click the button"
```
I'd guess that the same would carry for *numbers*, so what if we tweak the *autoconvert* to something like this:
``` swift
let myDouble = myInt.convert(.Double)
let myFloat = myInt.convert(.Float)
let myBool = myInt.convert(.Bool)
```

goktugyil:
That would be the same as myInt.convertToDouble

lfarah:
@goktugyil makes sense....

amosavian:
I prefer auto() rather than autoConvert()

piv199:
> This looks cool and useful, but how about type to type conversions? Double to Int? String to Int? etc.. Would we have to add a similar functionality to every types?

Yes, we should and would

> Can we create a general master "Number" type? So autoConvert() changes the thing, what ever it is to the general number and then changes back to whatever needed automatically?

How? Do we need to have some kind of NSNumber for Swift? But with auto() or value() or number()?

Some examples
```
let doubleNum = 3.0
let number = Number(doubleNum) // or doubleNum.toNumber() or doubleNum.number
let double: Double = number.auto()
func test(var1: Int) { ... }
test(number.auto())
```
Hovewer bad:
```
let doubleValue = 3.0
test(doubleValue.number.auto())
```
or should we have something
```
let doubleVal = 3.0
let intVal: Int = doubleVal.auto()
test(doubleVal.auto())
```

goktugyil:
I'm closing this. While its a cool trick, it causes more confusion than its worth.

lfarah:
makes sense

PR #93, opened 2016-03-02
Added removeSubview to UIView

### LINKED ISSUE #90: UIView.removeSubview(x,y)
Normally, you can only remove a subview if you have the _object_. In this method, you can remove a subview by having it's **x** and **y**
What do you guys think?

``` swift
extension UIView {
  func removeSubview(x:Double,y:Double) {
    for view in self.subviews {
      if Float(view.frame.origin.x) == Float(x) && Float(view.frame.origin.y) == Float(y) {
        view.removeFromSuperview()
        "removed"
      }
    }
  }
}
```

lfarah:
fixes #90

### DIFF
--- a/Sources/UIViewExtensions.swift
+++ b/Sources/UIViewExtensions.swift
@@ -229,6 +229,16 @@ extension UIView {
         }
     }
     
+    //EZSE: Remove view's subview from it's x and y locations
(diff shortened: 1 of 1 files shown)


goktugyil:
What does the  "removed" text do?

PR #102, opened 2016-03-22
Add missing tests for Array extensions. Make each public again with t…

pixel4:
Hey man,
I've added tests for the Array extensions. Both `each` methods are now testable so you can see how they work and we can (hopefully) bring them back as `public` to be of use. They're probably the most used methods for us. Type inference is really the key here, making the intend clear with less typing ;)

- Add missing tests for Array extensions. 
- Make each public again with tests. 
- Remove use of deprecated method.

I'll work on the Dictionary extensions next.

### DIFF
--- a/EZSwiftExtensionsTests/EZSwiftExtensionsTestsArray.swift
+++ b/EZSwiftExtensionsTests/EZSwiftExtensionsTestsArray.swift
@@ -63,4 +63,71 @@ class EZSwiftExtensionsTestsArray: XCTestCase {
         XCTAssertNotNil(numberArray.random())
         XCTAssertNil([].random())
     }
+
+	func testTestAll() {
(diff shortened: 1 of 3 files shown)


goktugyil:
Thanks for the PR :+1: 
What was the problem with requestURL?

pixel4:
No problems.

`NSURLConnection -> sendSynchronousRequest:returningResponse:error` is deprecated on iOS9 and `NSURLSession.sharedSession().dataTaskWithRequest` has been available since iOS 7 :) 

https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSURLConnection_Class/index.html#//apple_ref/occ/clm/NSURLConnection/sendSynchronousRequest:returningResponse:error:

pixel4:
Thanks for merging the changes, you're the man!

PR #107, opened 2016-03-29
Style fixes

inket:
Major (breaking) changes:
- Added [SwiftLint](https://github.com/realm/SwiftLint)
- Removed the force-casting and force-unwrapping, which means some method signatures had to change:
  - `Dictionary.random() -> NSObject` is now `Dictionary.random() -> Value`
  - `ez.appDisplayName() -> String` is now `ez.appDisplayName() -> String?`
  - `ez.appVersion() -> String` is now `ez.appVersion() -> String?`
  - `ez.appBuild() -> String` is now `ez.appBuild() -> String?`
  - `BlockButton.init(coder...)` is now `BlockButton.init?(coder...)`
  - `BlockWebView.init(coder...)` is now `BlockWebView.init?(coder...)`
  - `UIDevice.idForVendor() -> String` is now `UIDevice.idForVendor() -> String?`
  - `UIImage.compressImage(rate: ...) -> NSData` is now `UIImage.compressImage(rate: ...) -> NSData?`

My reasoning behind removing force-casting and force-unwrapping is that:
- It's just a matter of time before any of those start crashing. Optionals are optionals for a reason, and we shouldn't ignore the case where they may not meet our expectations.
- It lays the responsibility on EZSwiftExtensions for crashes caused by that. It's better to leave that responsibility to the user.

If this looks alright for you and you would like to merge it, inform me so I can also update the readme and documentation as necessary.


goktugyil:
Wow lots of changes!

I'll take a look at these later. 

Making big, multifile changes like this is not the best practice, making individual pull requests for each file or main feature is recommended.

goktugyil:
And were you receiving crashes while using this or did you just want to de-risk it?

inket:
> Making big, multifile changes like this is not the best practice, making individual pull requests for each file or main feature is recommended.

True. I was thinking they all fall under "style fixes", and I did separate the changes by commit. Hopefully the new GitHub tools make it easier to handle.

> And were you receiving crashes while using this or did you just want to de-risk it?

It's just de-risking. Also, SwiftLint's rules prohibit force-casting.

goktugyil:
Don't you think its better to add SwittLint to the example project and not the framework itsself? 

Why did you edit the appdelegate to conform to Lint while its not active in the example project?

Why were a part of this class got indented and the other part did not? Was the indentation a mistake also all those white space changes? https://github.com/inket/EZSwiftExtensions/blob/style-fixes/EZSwiftExtensionsTests/EZSwiftExtensionsTestsArray.swift

The random whitespace changes are really making it hard to read the real changes. Definitely should've been another commit

Anyways the changes make sense, I'm merging them. Going to edit some parts in my next commits. Thanks alot!

Please wait for my push, then feel free to update docs and readme, but separately :grin:

inket:
> Don't you think its better to add SwittLint to the example project and not the framework itsself?

But then you would have to build the example project for SwiftLint to trigger, which some people might not do. With the current settings it also checks the example project's code, but it's only triggered by building the framework (which the example project also does iirc)

> Why did you edit the appdelegate to conform to Lint while its not active in the example project?

Lint warning because the comments were too long. Like I said, SwiftLint checks the example project's code too, even when only building the framework.

> Why were a part of this class got indented and the other part did not? Was the indentation a mistake?

Oops. When I looked through the file with Sublime Text (with invisible characters visible) I saw that it was partly indented with tabs (and not spaces like the rest of the project). I must have messed up when searching & replacing the tabs with spaces.

> The random whitespace changes are really making it hard to read the real changes. Definitely should've been another commit

And they mostly were. 😜

Thanks for merging.

goktugyil:
I did my changes, feel free to edit now.

https://github.com/goktugyil/EZSwiftExtensions/blob/master/CONTRIBUTING.md
Yes enforcing style is a good idea, can you add the enforcements in your mind here? 

I'm also starting to think whitespace is not worth it, such a pain and I don't see the gain of it. Am I missing something?

PR #112, opened 2016-04-08
Add convenience method for a dismiss button to UITextView extension

dennis-tra:
With this method it's possible to add a toolbar with a done button to UITextView's keyboard with one line of code. 

Example:
```
class ViewController: UIViewController {
    @IBOutlet var textView: UITextView!

    override func viewDidLoad() {
        super.viewDidLoad()

        textView.addDoneButton()
    }
}
```

simulator screen shot 08 apr 2016 23 01 37

I used this little extension in a lot of projects by now and thought of sharing it.

### DIFF
--- a/Sources/UITextViewExtensions.swift
+++ b/Sources/UITextViewExtensions.swift
@@ -27,4 +27,20 @@ extension UITextView {
         editable = false
         scrollEnabled = false
     }
+    
+    /// EZSwiftExtensions: Automatically adds a toolbar with a done button to the top of the keyboard. 
+    /// Tapping the button will dismiss the keyboard.
+    func addDoneButton(barStyle: UIBarStyle = .Default) {
+        
+        let keyboardToolbar = UIToolbar()
+        keyboardToolbar.items = [
+            UIBarButtonItem(barButtonSystemItem: .FlexibleSpace, target: nil, action: nil),
+            UIBarButtonItem(title: "Done", style: .Done, target: self, action: #selector(resignFirstResponder))
+        ]
+        
+        keyboardToolbar.barStyle = barStyle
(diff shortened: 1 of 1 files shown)


goktugyil:
What is the done button used for?

dennis-tra:
TL;DR: It calls resignFirstResponder on the textView and dismisses the keyboard. 

Unlike for textFields for textViews the return button on the keyboard is often used to add a new line to the text instead of dismissing the keyboard. One way would be to implement a tapGestureRecognizer to dismiss the keyboard when the user taps outside of the textview on the viewController. Since textViews commonly display more text than usual it takes up much of the screens space and make it hard for the user to find a place to tap and dismiss the keyboard (especially on smaller screens). So we added this done button to the toolbar.

_Side note: In case the textView is embedded in a tableViewCell the tableView is scrolled to the perfect position, so that the keyboard + toolbar don't cover anything of the textView._

goktugyil:
Wouldn't the "Done" text change in each scenario? Most of the ones I see are usually "Send"? Maybe the user might want to send the text as a variable

dennis-tra:
It's easy to change the extension to something which allows:
`textView.addDoneButton(title: "Title")`
Changing this button to "Send" (or any other imperative verb) implies, that tapping the button follows an action (which is actually not what I thought of). It was just meant for dismissing the keyboard to continue interacting with other UI elements on the same screen.

goktugyil:
I see, we might need that additionally too. With a closure or something to call after the action

goktugyil:
Did it https://github.com/goktugyil/EZSwiftExtensions/commit/d17f6d141f22fcdc02dbf9a823bf03b5fc54aba2

PR #118, opened 2016-04-18
Added extensions for ceil in double.

arasu01:


goktugyil:
Thanks 👍 
Mind adding a test case for this?

arasu01:
Thanks for Merging. 
I will add the test case early next.

PR #122, opened 2016-05-13
Swift NSDate extension with nullable convenience init crashes EXC_BAD_ACCESS

piv199:
Just got a situation when user enters date via string in my application, then I pass entered string in NSDate(fromString:format:) and estimate to get nil if it isn't valid, however I got EXC_BAD_ACCESS. I went to stackoverflow and found that it's a problem for new xcode 7.3.

http://stackoverflow.com/questions/36189714/new-xcode-7-3-swift-nsdate-extension-with-nullable-convenience-init-crashes-exc

I also added a test case for creating NSDate from invalid string. I don't know should we have one now, but if we had, then we could find this error earlier.


goktugyil:
Thats a silly bug. But thanks for the pull request!

PR #123, opened 2016-05-14
Fixed access control for NSAttributedString

GabrielCartier:
NSattributedString extension has internal functions and can't be used on outside projects.

Made the methods public.


goktugyil:
Classic mistake 😄  thanks

PR #131, opened 2016-05-27
Some new extensions for String, NSDate, NSURL, NSDictionary, UIApplication, UIImage, Double/Float

amosavian:
For String:
- Base64 encoding/decoding
- Regular expression matching
- Counting words and paragraphs
- Coloring substring
- Get position of substring
- pattern match for hasPrefix() and hasSuffix to use in select-case
- Height of rendered text in view

Double/Float:
- String initializer with precision formatting

NSURL:
- Get file size from remote server and checks if it support resume
- File (local) related function : get size, modification date and creation date
- Checks if path is a directory

NSDate:
- Init NSDate from HTTP response header date

NSDictionary:
- JSON serialization 

UIApplication:
- run a block after app resigns activity in iOS

UIImage:
- colorized pattern from image

### DIFF
--- a/EZSwiftExtensions.xcodeproj/project.pbxproj
+++ b/EZSwiftExtensions.xcodeproj/project.pbxproj
@@ -17,6 +17,14 @@
 		7605D3171C81FC180046FAC3 /* EZSwiftExtensionsTestsInt.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7605D3071C81FBB80046FAC3 /* EZSwiftExtensionsTestsInt.swift */; };
 		7605D3181C81FC180046FAC3 /* EZSwiftExtensionsTestsNSDate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7605D3081C81FBB80046FAC3 /* EZSwiftExtensionsTestsNSDate.swift */; };
 		7605D3191C81FC180046FAC3 /* EZSwiftExtensionsTestsString.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7605D3091C81FBB80046FAC3 /* EZSwiftExtensionsTestsString.swift */; };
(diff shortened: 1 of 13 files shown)


goktugyil:
Thanks for the pull request. These seem like potentially dangerous extensions. Can you add some tests to these additions?

Also +    /// EZSE: Returns remote size of url, don't use it in main thread
Its better to handle threads inside the function, but then there is the return problem. hmm

Also just adding files doesn't work. They have to be referenced in both targets, 

Like this: https://github.com/goktugyil/EZSwiftExtensions/commit/50e2511dfd8246e125516bf3015c107fe8e7a493

The files need to be in the sources folder. I create the files in somewhere else then, add the files in the sources folder from finder. Then drag the new file inside the sources folder to the 2 targets inside the xcode project and place them in the sources folder in there. Copy file prompt is unchecked. I hope this makes sense. @lfarah

amosavian:
hey @goktugyil ,
They are variables with return values. we can replace them with functions with completion handlers in this case. Though that NSLog notification were enough for me, you may decide it's not ideal for general, or replace NSLog with assert. Tell me what you prefer.
About dropping files inside project I created files here and have no clone on my desktop. I will clone and try to update project files and paste them here again

goktugyil:
Well multi thread code is annoying, except when you use ez.runthisinbackground 😁 
Completion handlers seems like a better idea.

amosavian:
hi @goktugyil ,
please check these new commits

goktugyil:
Seems good. All thats missing seems to be tests

amosavian:
@goktugyil now test has been implemented and some minor syntax errors has been fixed

goktugyil:
Now we have merge conflicts :(

amosavian:
How about closing this pull and I create a new one?

loveuqian:
@goktugyil 
want a new extension for image to crop circle image

goktugyil:
@loveuqian what do you mean? There is a UIView.roundSquareImage & .roundCorners

PR #139, opened 2016-06-04
Double, String, NSDate, NSURL, NSDictionary, UIApplication and UIImag…

amosavian:
…e Extensions and Tests

Double:
- String inits with precision
NSDate:
- init from HTTP date strings
NSDictionary:
- JSON serialize and deserialize
NSURL:
- file properties methods
- get remote server size of URL
- checks server supports resuming
- comparing URLs
String:
- determining position of substring
- counting words and paragraphs
- get matches of regular expression
- get height of rendered text (usable in determining UILabel/UITextView height programatically
- colorize instances of a substring inside string
- hasPrefix and hasSuffix pattern matching
UIApplication:
- convenience method to run a block when app is in background
UIImage:
- colorize uiimage (pattern colouring)


goktugyil:
Thanks 👍

PR #151, opened 2016-07-02
UIStoryboard extension for easily instantiating view controllers

piv199:
Creating view controllers from code using storyboards is a lot of boilerplate.

From
let vcName = ViewController.className;
let vc = storyboard!.instantiateViewControllerWithIdentifier(vcName) as! ViewController

To
let vc = storyboard!.instantiateVC(ViewController)

Both vc are of type ViewController.


goktugyil:
Thanks 👍

PR #157, opened 2016-07-06
Added capitalizeFirst() and getCapitalizedFirst() plus tests

### LINKED ISSUE #150: capitalizeFirst should be a method
```capitalizeFirst``` inside ```StringExtensions``` realizes an action of capitalizing the first character of a string, so therefore it should be a method instead of a variable with a getter.

Today's implementation:
``` swift
	/// EZSE: Capitalizes first character of String
	public var capitalizeFirst: String {

		guard characters.count > 0 else { return self }
		var result = self

		result.replaceRange(startIndex...startIndex, with: String(self[startIndex]).capitalizedString)
		return result
	}
```

What it should be:

``` swift
	/// EZSE: Capitalizes first character of String
	public func capitalizeFirst() -> String {

		guard characters.count > 0 else { return self }
		var result = self

		result.replaceRange(startIndex...startIndex, with: String(self[startIndex]).capitalizedString)
		return result
	}
```

lfarah:
fixes #150

### DIFF
--- a/EZSwiftExtensionsTests/EZSwiftExtensionsTestsString.swift
+++ b/EZSwiftExtensionsTests/EZSwiftExtensionsTestsString.swift
@@ -10,110 +10,116 @@ import XCTest
 @testable import EZSwiftExtensions
 
 class EZSwiftExtensionsTestsString: XCTestCase {
-    var string: String!
+	var string: String!
 
-    override func setUp() {
-        super.setUp()
-        string = "0123456789"
-    }
+	override func setUp() {
+		super.setUp()
+		string = "0123456789"
+	}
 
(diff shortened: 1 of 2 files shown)


piv199:
Why do you duplicate function behaviour, capitalizeFirst, probably, should depend on capitalizedFirst?
You missplaced functions getCapitalizedFirst should not be mutating, it should return new string
Please, consider naming like, capitalizeFirst() and capitalizedFirst(), the first one is mutating

goktugyil:
I agree with piv199. It should be the opposite.

lfarah:
This was done on #160. Closing

PR #159, opened 2016-07-18
Change "panned" to "pinched" in print call

rpural:
In an example of the mainview.addPinchGesture(target: self, action: "userAction") call, the print() statement had been copied down from the addPanGesture() call above, and not changed to reflect the new functionality.


goktugyil:
Thanks 👍

PR #165, opened 2016-08-03
capitalize(d)Prefix + capitalize(d)Suffix + tests

piv199:
Wrapped this ugly thing:

```
str.replaceRange(startIndex..<startIndex.advancedBy(min(count, length)),
                          with: String(self[startIndex..<startIndex.advancedBy(min(count, length))]).uppercaseString)
```

into this:

`str.capitalizeFirst(count)`

Also this PR includes fixes for SwiftLint (legacy constructors)


piv199:
Hm... just realised that methods should be named uppercase(d)Prefix and uppercase(d)Suffix.

piv199:
Just found an interesting thing

`string[0..<3].uppercaseString` is valid. 
But it's not ensures whether string has index of 0,1,2 or not. Thus it's could lead to crash.

`string[0..<min(3, string.length)].uppercaseString` vs `string.uppercasedPrefix(3)`

goktugyil:
Awesome 👍

PR #167, opened 2016-08-06
Lowercase(d)First, lowercase(d)Prefix, lowercase(d)Suffix, lowercase(d)(range:) + tests

piv199:
Added counter methods for uppercases.

I'll update Readme.MD with all this methods in next pull requests.


goktugyil:
Thanks 👍

PR #170, opened 2016-08-17
add multiple subviews in UIView

Yu-w:
As titled: added function that can add multiple subviews in a UIView instance.


goktugyil:
Thanks 👍

PR #171, opened 2016-08-17
add decompose function in Array

Yu-w:
Function that decompose an array to a tuple with first element and the rest; useful in Functional Programming.
For example, doing a quicksort in a few lines:
```swift
extension Array where Element: Comparable {
    func quicksort() -> Array<Element> {
        if let (head, tail) = self.decompose() {
            return tail.filter { $0 <= head }.quicksort() + [head] + tail.filter { $0 > head }.quicksort()
        } else { return [] }
    }
}
```


goktugyil:
Can you also write a test case for this?

piv199:
Is it okay that decompose() returns nil on empty array? Is it supposed behaviour?

Yu-w:
Of course. Decompose basically immutably extract first element from the array and the <optional> return type guarantee the correctness and safety of using this function.

Yu-w:
Basically the features of optional type

PR #172, opened 2016-08-17
deprecate each in Array Extension

Yu-w:
Deprecate `each` function in Array extension, because there's native `forEach` function that serves exactly the same as `each`. Feel no need for duplicating same method.


goktugyil:
Thanks 👍

PR #173, opened 2016-08-17
GCD Extension

Yu-w:
Add dispatch after function to make a delay, which should be the standard way to making delay over a closure. http://stackoverflow.com/questions/27990085/nstimer-how-to-delay-in-swift

Add helpers to access global queue easily.


goktugyil:
Thought about adding those shortcuts myself a while ago. Thanks for doing it 👍

PR #174, opened 2016-08-17
deprecate `mapFilter` function in Array Extension

Yu-w:
Deprecate `mapFilter` because there's native function`flatMap` that does the same thing. Feel no need to have duplicated methods.


goktugyil:
Thanks 👍

PR #175, opened 2016-08-19
Few refinement and new algorithm for computing top most view controller

Yu-w:
As titled. Thanks!


piv199:
What the purposes of changing `topMostVC` to `topMostViewController`. This is not only one method in this library that has suffix VC as for ViewController. 

Especially, I love EZ for it's simplicity of naming.

PS. also in isBlank tests you are doing the same verification twice.

Yu-w:
1. Changing name is for following Apple's naming convention and reducing ambiguity especially for beginers.
2. If there are still some method name with VC in the lib, we should also consider change all of them to proper naming convention.
3. Nice catch for isBlank!

goktugyil:
I don't think any beginner at that level (Not knowing what VC stands for) will be using this library.

goktugyil:
Thanks 👍

PR #177, opened 2016-08-20
New algorithm for UIImage Extension methods

Yu-w:
Using `UIBezierPath` to `addClip` UIImage to have roundedCorner is better than simple setting corner radius in two ways:
1. Avoiding side effects of setting additional `clipToBounds`
2. Avoiding offscreen rendering; useful when displaying a bunch rounded images

New class explicit for image downloading; a lot efficient and thread-safe than simple url request.


piv199:
roundSquareImage it's really cool! Could we parameterize it with level of corning?

Yu-w:
Yeah, maybe another pull request later

goktugyil:
What kind of test cases would be appropriate for this?

Yu-w:
To be honest, have no idea how to test yet... I've used this two in my projects a lot and haven't found a single problem yet.

lfarah:
@goktugyil can we merge this without tests?
@Yu-w is this working in Swift 3?

goktugyil:
I don't think it will work in Swift3, it has NSURL stuff. 

It also seems to have project reference issues. I can fix that easily after Swift3 problems are fixed.

lfarah:
Guys, now we can work on this since 3.0 was approved

Issue #110, opened 2016-04-07
Wrong addAsChildViewController's calls order


respan:
Hey. According to [Apple docs](https://developer.apple.com/library/ios/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html#//apple_ref/doc/uid/TP40007457-CH11-SW1) (section "Adding a Child View Controller to Your Content") you should first call `addChildViewController` and only then `addSubview`:

```
    /// now
    public func addAsChildViewController(vc: UIViewController, toView: UIView){
        toView.addSubview(vc.view)
        self.addChildViewController(vc)
        vc.didMoveToParentViewController(self)
    }

    /// how it should be
    public func addAsChildViewController(vc: UIViewController, toView: UIView){
        self.addChildViewController(vc)
        toView.addSubview(vc.view)
        vc.didMoveToParentViewController(self)
    }
```
I didn't experience any bugs, but still have some concerns. Should i submit PR?

goktugyil:
There was some kind of issue before about the view, that only happened here. I am not sure if this was the problem but something about the order messed up when I tried to access the size of view or something like that. Feel free to experiment, I'll also check it out again.

I seem to have used the correct way in my other repos:
https://github.com/goktugyil/EZSwipeController/blob/master/EZSwipeController.swift#L177-L179

goktugyil:
https://pspdfkit.com/images/blog/2016/swift-extensions-0ec098cd.png
Here is something similar I came across to

lfarah:
@respan I'd be cool to submit a PR. It's always good to follow the Apple Docs

goktugyil:
If anything breaks lets keep in mind that it was the order.

PR #179, opened 2016-08-20
update addAsChildViewController's calls order

Yu-w:
#110


goktugyil:
If anything breaks lets keep in mind that it was the order.

PR #185, opened 2016-08-25
Error type tests

piv199:
toString should not be a property, it should be func, or property like string.

Also, current behaviour is not as expected:
- [ ] enums return its case representations to strings, but `case CaseOne = "some case"` still returns `"CaseOne"`, don't know for this case, how it should be handled
- [ ] structs return `StructTypeName()` instead `StructTypeName`
- [ ] classes return `Module.ClassName` instead `ClassName`

### DIFF
--- a/EZSwiftExtensionsTests/ErrorTypeTests.swift
+++ b/EZSwiftExtensionsTests/ErrorTypeTests.swift
@@ -8,20 +8,20 @@
 
 import XCTest
 
-enum CustomErrorEnum: ErrorType {
-    case CustomError, AnotherError
+enum ErrorEnum: String, ErrorType {
+    case CustomError
+    case AnotherError = "String Another Error"
 }
 
-struct CustomErrorStruct: ErrorType { }
+struct ErrorStruct: ErrorType { }
+
+class ErrorClass: ErrorType { }
 
 class CustomErrorClass: ErrorType {
-
     let description: String
-
     init(description: String) {
         self.description = description
     }
-
     var toString: String {
         return description
     }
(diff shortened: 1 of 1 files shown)


goktugyil:
Why does it matter if its var or func? They are both funcs in the end of day. What do you mean by property like string?

piv199:
@goktugyil I mean that `toString` it reads like convert to string, make to string, so its an action it should be func. Whether, for example `error.string` or `error.stringValue` it's a property.

goktugyil:
Oh, I thought you were mentioning the other toStrings. Yeah for error it makes sense. stringValue look better imo.

goktugyil:
ErrorType.toString is intended for debugging and error logging reasons. 

So in my main project, I have code like this.
When I try to print out the error it prints something like this:

`EZSwiftExtensions.MyFirstClass.MySecondClass.MyError.Problem3
`
I wanted it to print the raw error without writing custom descriptions. Like only "Problem3"

```
public class MyFirstClass {
    class MySecondClass {
        enum MyError: ErrorType {
            case Problem1
            case Problem2
            case Problem3
        }
        static func getError() -> MyError {
            return MyError.Problem3
        }
    }
}
```
I'm also trying in the library project and I'm not getting the same results as my project :/

Apple might have changed this, also we need to try a few Apple framework errors to see how it plays with them.

goktugyil:
If there isn't a good use case for this then I'm open to deleting this feature. 

If there are other use cases with errortype, its a good time to add them before deleting the files.

piv199:
@goktugyil why not having for debugging features?

Also maybe it makes sense `EZSwiftExtensions.MyFirstClass.MySecondClass.MyError.Problem3`, doesn't it? What if there are `EZSwiftExtensions.MyFirstClass.MySecondClass.MyError2.Problem3` and `EZSwiftExtensions.MyError.Problem3`.

And printing Problem3 isn't enough for understanding what actually type was it and where it came from.

However as average developer I possibly better get `Problem3`. It's a bit nicer and cleaner. But for what purposes and intents? Why do I need if not for debugging.

piv199:
Okay another case

```
enum ErrorIntEnum: Int, ErrorType {
    case Error0 = 0
    case Error2 = 1
}
```

```
enum ErrorStringEnum: String, ErrorType {
    case Error0
    case Error2 = "Error1"
}
```

print(ErrorIntEnum(rawValue: 1)!.toString)
print(ErrorStringEnum(rawValue: "Error1")!.toString) 

What's the intent of user? Does he want to get rawValue or the string value of case if it's enum?

piv199:
I actually haven't used it in my projects, so anybody provide with a real use case for these methods, so we could make this method to work as it should. Otherwise it this will be closed and Error Type extension will be deleted.

goktugyil:
Tested a bit, tried the  cases you suggested, I'm removing the current extension but not deleting the file in case someone will find a good use case.

PR #188, opened 2016-09-02
UIStoryboardTests + added necessary test files + added initialVC + tests

piv199:
Okay some questions again :)

- [ ] Testing mainStoryboard requires NSBundle(ofClass: self.dynamicType) in XCTest. Could we achieve this?
- [ ] `instantiateVC(UnknownType)` produces ObjC exception if there is no such storyboard ID found in storyboard. Also currently `instantiateVC(T.Type)` returns optional, so it's a bit weird - Somebody could think that if there is no such storyboard ID than it will return nil. 

Please revise structure and maybe I missed something or done in wrong way...
Help appreciated. Thanks.


lfarah:
I think @goktugyil will be better than me on answering this question

piv199:
I redo it for swift 3.0 after release version.

goktugyil:
👍

Issue #205, opened 2016-09-25
keyboardWillShowWithFrame and similar should be marked as open


AdamSliwakowski:
Right now there is no possibility to override `keyboardWillShowWithFrame` because it is marked only as `public`.

piv199:
@AdamSliwakowski Thanks for opening issue, are there any related `public` instead of `open` functions?

lfarah:
@piv199 can we close this?

piv199:
@lfarah Not yet

piv199:
Wow, I missed this issue, should be done in 1.8 or 1.8.1

lfarah:
Just marked as 1.8.1

PR #206, opened 2016-09-25
open vs public issues #205

piv199:
This is fast fix for #205. There are could be some more methods that requires `open` instead of `public`.


cherishloveyou:
please merge to master

Khalian:
@cherishloveyou https://github.com/goktugyil/EZSwiftExtensions/pull/399.

PR #218, opened 2016-09-29
FloatingPointExtensions + tests

piv199:
Swift 3.0 API design guidelines:

`getRoundedByPlaces(_:)` is now `rounded(toPlaces:)`
`getCeiledByPlaces(_:)` is now `ceiled(toPlaces:)`
`roundByPlaces(_:)` is now `round(toPlaces:)`
`ceilByPlaces(_:)` is now `ceil(toPlaces:)`

**Significant change** Methods above are now part of FloatingPoint protocol extension with default implementation.

**Motivation:** 

round(ed) and ceil(ed) are useful for all floating types. In 1.7 version we have Double implemented round(ed) and ceil(ed), but Float are still missing these methods.

Okay, we could just copy-paste methods to Float extension, but it is not good to ignore reusability.

Let's move to protocol representing floating types - FloatingPoint protocol.


piv199:
Do not merge, need to verify results from

http://codereview.stackexchange.com/questions/142748/swift-floatingpoint-rounded-to-places/142850#142850

piv199:
#229

lfarah:
@goktugyil I'd like you to approve this once since it is changing a lot of files....

goktugyil:
Looks good. Have to run it to make sure ofc

lfarah:
1. Why is it better to extend the tests than to write the method inside the class.
2. @goktugyil You said in #225 that you didn't want global variables. Should we change in here?

goktugyil:
Yeah. Could also be done after merge too

Issue #217, opened 2016-09-29
imageWithUrl(url: String) not update imageView instantly


DesmondPang:
It needs to assign the image in main thread otherwise it will not update the imageview. 

```
 /// EZSwiftExtensions
    public func imageWithUrl(url: String) {
        ez.requestImage(url, success: { (image) -> Void in
           
            if let img = image {
                DispatchQueue.main.async {
                    self.image = img
                }
            }
        })
    }
```

piv199:
Done, will be available in v1.8

PR #220, opened 2016-09-29
fixed bug setting image on undefined thread for `imageWithURL` related methods

piv199:
Also Removed code duplicate

#217

### DIFF
--- a/Sources/UIImageViewExtensions.swift
+++ b/Sources/UIImageViewExtensions.swift
@@ -73,7 +73,9 @@ extension UIImageView {
     public func imageWithUrl(url: String) {
         ez.requestImage(url, success: { (image) -> Void in
             if let img = image {
-                self.image = img
+                DispatchQueue.main.async {
+                    self.image = img
+                }
             }
         })
     }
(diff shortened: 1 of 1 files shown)


lfarah:
Hmmmm, I read the method ```imageWithUrl(url:)``` and I'm thinking about how this naming can be better in Swift 3. Maybe do an initializer like ```UIImageView(url: String)```

What are your thoughts?

piv199:
@lfarah looks awesome, but what's about supporting storyboards?

lfarah:
I didn't see any @IBDesignable or @IBInspectable... How is this extension supporting storyboards?

piv199:
@lfarah imagine you set up storyboard view controller with view and UIImageView on it, then you set reference to this UIImageView in your ViewController class. Once ViewController viewDidLoad you would like to load an image into UIImageView from url.

You cannot reassign `UIImageView` with `UIImageView(url: String)`, but you can do `imageView.imageWithURL(...)`

lfarah:
Alright. Now I understood what you meant by "supporting storyboards".
What if we add this initializer and also change the name to something like:```imageView.image(urlString:...)```, so it looks more like Swift 3

piv199:
@lfarah yes it makes sense, also I think there are should be support for urlString and url. About naming, yes, I haven't changed this file yet

piv199:
@lfarah this is just hot fix, naming is about another pull request, also adding new features is out of scope current pr

Issue #221, opened 2016-09-29
Add Danger.systems


lfarah:
We should use [Danger](http://danger.systems/) to automatically fail a pull request if it doesn't include tests, as well as a bunch of other stuff.

piv199:
@lfarah https://github.com/fastlane/fastlane/blob/master/Dangerfile

as example

goktugyil:
Done?

lfarah:
Yup.

PR #223, opened 2016-09-30
Travis ci strikes back

piv199:
xcodebuild instead of xctool
Two schemes: iOS and tvOS
Also added pod lint

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,26 @@
 osx_image: xcode8
 language: objective-c
-rvm: 2.2.2
 cache: cocoapods
 
 env:
   global:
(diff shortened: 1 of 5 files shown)


piv199:
Travis not strikes back, still need to fix iOS scheme, tvOS testing is working

lfarah:
Damn.... do you even sleep? hahahahah

piv199:
@lfarah yes, I'll create a new pull request that more accurate shows my changes to start work travis.
Now it really seems to do what it should, also codecov is working 👍

lfarah:
#221 is a little bit harder but I'll try to do it over the weekend

PR #225, opened 2016-10-01
Added Character Tests

lfarah:


goktugyil:
This isn't making any sense to me.

I also like tests having their own variables inside functions, instead of setting up or having class level variables. Its easier to separate tests from each other.

lfarah:
@goktugyil you want me to revert on how it was on the first commit?

goktugyil:
Yeah I think that was more manageable. Think about it, if 3 people are writing tests for the same func, they're not really going to know or case about the variables. Sometimes one variable is going to need changing, but that might break the other tests.

piv199:
@goktugyil well I could agree about visibility, under stability, but not about changing and breaking tests. Tests are independent, and global variables are always reinitialized for each test.

Issue #222, opened 2016-09-29
Add codecov


lfarah:
We should integrate [codecov](https://codecov.io/) to check how's the code coverage on each pull request

lfarah:
Yay, it works! 🎉 
<img width="807" alt="screen shot 2016-10-01 at 12 57 45 am" src="https://cloud.githubusercontent.com/assets/6511079/19012710/3ee82f72-8772-11e6-990f-24c39fe80f13.png">

PR #226, opened 2016-10-01
Travis ci strikes back back

piv199:
xcodebuild instead of xctool
Two targets instead only iOS
xcpretty shows nice output
Added pod lib lint
codecov by @lfarah #222


lfarah:
👍👍👍👍👍

PR #241, opened 2016-10-07
Danger

piv199:
Restrict changing only one test/extension per PR
Fail travis on danger error


lfarah:
I really liked your idea!!! Danger is all about being creative!

PR #243, opened 2016-10-09
<Description> : Adding urlEncoding of strings extension and test for the same

Khalian:
.<Description> : Adding urlEncoding of strings extension and test for the same

<Type> : feature/test


goktugyil:
@lfarah  why has this failed 😁

piv199:
@goktugyil there were 3 runs on travis for this pull request, two failed and one is succeeded (the last one). But it's really interesting, why it left this message and why were 3 runs?

PR #244, opened 2016-10-09
<Description> : Adding urlEncoding of strings extension and test for the same

Khalian:
<Description> : Adding urlEncoding of strings extension and test for the same.

<Type> : feature/test

### DIFF
--- a/Sources/StringExtensions.swift
+++ b/Sources/StringExtensions.swift
@@ -452,7 +452,11 @@ extension String {
     }
     
     #endif
-
+    
+    // EZSE: URL encode a string (percent encoding special chars)
+    public func urlEncode() -> String {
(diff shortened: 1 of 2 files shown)


Sources/StringExtensions.swift
```diff
@@ -452,7 +452,11 @@ extension String {
     }
     
     #endif
-
+    
+    // EZSE: URL encode a string (percent encoding special chars)
+    public func urlEncode() -> String {
```
piv199: Should be urlEncoded, urlEncode otherwise mutate the string.
Refer ro array.sort(ed)

piv199:
Hi, thanks for the pull request, don't mind if danger fails. 
We are still configuring it, so just make change to urlEncoded and pull request can be merged.
If you would like you could also make a mutating version urlEncode.

Khalian:
Ya sure I can do that.

Issue #242, opened 2016-10-07
New Danger tests


lfarah:
so now we need to find a way to
* Ensure documentation in the method by checking for ///
* Ensure that each new PR of a new extension has a Test.

I have an idea on how to easily do number 2: We need to have a [Pull Request Template](https://github.com/vsouza/awesome-ios/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and if the user checked ```[X]``` for ```New extension```, we ensure tests

piv199:
My drafts
- On added file count or check modifying pbproject file
- Big PR, total lines out of 150-200 or more?
- If sources modyfied and changed symbols more than x, change changelog, or check PR if its #trivial
- swiftlint errors
- Work in progress PR
- forbid changing some files or include some text in PR if not a collaborator
- if PR is not mergable, warn, but it may be obvious by github

lfarah:
[Danger Xcode Summary Plugin](https://github.com/diogot/danger-xcode_summary) - A Danger plugin that shows all build errors, warnings and unit tests results generated from xcodebuild.

piv199:
On weekend will add support for trivial change and related changes

PR #248, opened 2016-10-12
Added Danger Xcode Summary Plugin from #242

lfarah:
I mentioned in #248 about it.
[Danger Xcode Summary Plugin](https://github.com/diogot/danger-xcode_summary) - A Danger plugin that shows all build errors, warnings and unit tests results generated from xcodebuild.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -17,3 +17,6 @@ end
 if git.modified_files.grep(/^EZSwiftExtensionsTests\//).count > 1
 	fail("Please, modify only one extension per pull request.")
 end
+
+xcode_summary.report 'xcodebuild.json'
+
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,4 @@
 source "https://rubygems.org"
 
 gem "danger"
+gem 'danger-xcode_summary'


piv199:
There is something you should add to xcodebuild to generate summary?

lfarah:
@piv199 hmmm, it doesn't say in the README....

@diogot can you help us?

diogot:
Hi @lfarah, sure!

@piv199 are correct. You need to generate the summary, maybe it's not very clear but it is on [README](https://github.com/diogot/danger-xcode_summary/blob/master/README.md) (we'll make a more detailed documentation)

> You need to use xcpretty with xcpretty-json-formatter to generate a JSON file that this plugin can read.

Our plugins need a json with the summary of issues and to create it we use `xcpretty`, here is an example on how to generate:

```
cat xcode-output.log | XCPRETTY_JSON_FILE_OUTPUT="xcodebuild.json" xcpretty -f `xcpretty-json-formatter`
```

In your case may be easy to generate on the build.

I create a pull request with a possible implementation #250

lfarah:
Hey @diogot, thanks for replying so fast and taking the time to actually open a PR.

I think it's worth saying something inside the usage like:

1. Install [XCPretty](https://github.com/supermarin/xcpretty)
2. Install [xcpretty-json-formatter](https://github.com/marcelofabri/xcpretty-json-formatter)
3. Just add this line to your Dangerfile:
``` swift
xcode_summary.report 'xcodebuild.json
```

PR #249, opened 2016-10-12
Fixes #228

### LINKED ISSUE #228: Timer.runThisEvery() should return a Timer instead of CFLoopTimer
Right now the ```Timer.runThisEvery()``` works like this:
``` swift
    public static func runThisEvery(seconds: TimeInterval, handler: @escaping (CFRunLoopTimer?) -> Void) -> Timer {
        let fireDate = CFAbsoluteTimeGetCurrent()
        let timer = CFRunLoopTimerCreateWithHandler(kCFAllocatorDefault, fireDate, seconds, 0, 0, handler)
        CFRunLoopAddTimer(CFRunLoopGetCurrent(), timer, CFRunLoopMode.commonModes)
        return timer!
    }
```

Because it is returning a ```CFRunLoopTimer```, we can't do timer.invalidate().

# How to fix it
Replace ```CFRunLoopTimer``` with ```Timer```

lfarah:
In #228 I mentioned the bug in TImer.runThisEvery().

### DIFF
--- a/Sources/NSTimerExtensions.swift
+++ b/Sources/NSTimerExtensions.swift
@@ -9,7 +9,7 @@ import UIKit
 
 extension Timer {
     /// EZSE: Runs every x seconds, to cancel use: timer.invalidate()
-    public static func runThisEvery(seconds: TimeInterval, handler: @escaping (CFRunLoopTimer?) -> Void) -> Timer {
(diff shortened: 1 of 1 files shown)


piv199:
Please rebase on #250

PR #252, opened 2016-10-17
<Description> : Adding tests for UI Switch methods.

Khalian:
<Type> : test


lfarah:
Interesting approach... I always thought it would take UI Tests to test the UI extensions, but that's not true since we can still test the value of them.

Khalian:
UI Tests for UI extensions would be integration tests. Since I was writing unit tests, I found it sufficient to test only the flags, state changes that the extensions are doing.

PR #253, opened 2016-10-18
<Description> : Adding UIUserInterfaceSizeClassTests.

Khalian:
<Type> : test


goktugyil:
Thanks 👍  I think if we could just extend enums and get their string value directly it would be more valuable. Kind of like this:

https://github.com/goktugyil/EZSwiftExtensions/commit/6edb4176c74c64bbc428a7b7698dbe1a70414ebf#diff-7beba89e7c564dc0e70633d5b2547b11

PR #254, opened 2016-10-18
<Description> : Adding a new array extension that returns only the unique elements in the array.

Khalian:
<Type> : feature/test

### DIFF
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -224,4 +224,19 @@ class ArrayTests: XCTestCase {
         XCTAssertTrue(numberArray.decompose()!.head == head)
         XCTAssertTrue(numberArray.decompose()!.tail == tail)
     }
+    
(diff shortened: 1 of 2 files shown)


lfarah:
Don't know why these NSURL Tests failed...
Really good idea for extension! I've used that last month on a small project and it would've been really handy if I had this

piv199:
@Khalian is this faster than `Array(Set(self))`?

Khalian:
I dont know, I did not benchmark it. In theory this should avoid the space complexity of the set, which is O(n). 

But honestly I do not know if Set(self) would actually do the same operation (a contains in the array before putting it into the set). So I cant make any theoretical justifications of time complexity. 

I admit though the Array(Set(self)) is much cleaner.

piv199:
@Khalian My approach is bad because it's not stable.

Khalian:
Is it not reasonable to admit an array element be hashable? Otherwise every contains call would become O(n).

PR #255, opened 2016-10-18
<Description> : Completing the CGFloat tests.

Khalian:
<Type> : test

### DIFF
--- a/EZSwiftExtensionsTests/CGFloatTests.swift
+++ b/EZSwiftExtensionsTests/CGFloatTests.swift
@@ -18,7 +18,17 @@ class CGFloatTests: XCTestCase {
         degrees = 360
     }
 
-    func testToggle() {
+    func testDegreesToRadians() {
         XCTAssertEqual(radians, degrees.degreesToRadians())
+    }
+    
+    func testDegreesToRadiansStaticVersion() {
+        XCTAssertEqual(CGFloat.degreesToRadians(degrees), radians)
+    }
+    
+    func testDegreesToRadiansMutableVersion() {
+        var mutableDegrees = CGFloat(360)
+        mutableDegrees.toRadiansInPlace()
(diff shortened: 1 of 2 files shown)


EZSwiftExtensionsTests/CGFloatTests.swift
```diff
@@ -18,7 +18,17 @@ class CGFloatTests: XCTestCase {
...
+    func testDegreesToRadians() {
         XCTAssertEqual(radians, degrees.degreesToRadians())
     }
+    
+    func testDegreesToRadiansStaticVersion() {
+        XCTAssertEqual(CGFloat.degreesToRadians(degrees), radians)
+    }
+    
+    func testDegreesToRadiansMutableVersion() {
+        var mutableDegrees = CGFloat(360)
```
piv199: `var mutableDegrees = degrees` - what do you think?

Sources/CGFloatExtensions.swift
```diff
@@ -14,13 +14,8 @@ extension CGFloat {
     public var center: CGFloat { return (self / 2) }
 
     /// EZSwiftExtensions
-    public func toRadians() -> CGFloat {
```
piv199: Why has it been removed?

Khalian: toRaidans does the exact same thing as degreesToRadians and they are both public methods. Its bad OOP design to have two public methods with identical functionality with one method calling the other(and not doing anything other than that).

piv199: @Khalian could you please mark it with deprecated and renamed to `degreesToRadians` as done in many other cases through `ArrayExtensions` and others...
As we move forward we will drop such methods, but give people a bit time to see that it will be removed soon.

lfarah:
Thanks @Khalian! 🎉

PR #256, opened 2016-10-19
Increased timeout to one minute

piv199:
This chunk of code sometimes fails in 30 seconds, increased to 60 seconds timeout


piv199:
Hm, possibly change url, cause it's not stable?

PR #260, opened 2016-10-20
Add a sub array function to Array and test

Dendim0n:
add two new functions:

    ///EZSE: Get a sub array from range of index
    public func subArray(fromIndex:Int,toIndex:Int) -> Array
    
    ///EZSE: Get a sub array from elements that meet specific condition
    public func subArray(_ condition: (Element) -> Bool) -> Array

### DIFF
--- a/Sources/ArrayExtensions.swift
+++ b/Sources/ArrayExtensions.swift
@@ -21,21 +21,12 @@ public func ==<T: Equatable>(lhs: [T]?, rhs: [T]?) -> Bool {
 extension Array {
     
     ///EZSE: Get a sub array from range of index
-    public func subArray(fromIndex:Int,toIndex:Int) -> Array {
+    public func get(at range:ClosedRange<Int>) -> Array {
         var subArray = Array()
-        for index in fromIndex...toIndex {
+        let lowerBound = range.lowerBound > 0 ? range.lowerBound : 0
+        let upperBound = range.upperBound > self.count - 1 ? self.count - 1 : range.upperBound
+        for index in lowerBound...upperBound {
             subArray.append(self[index])
-        }
-        return subArray
-    }
-    
-    ///EZSE: Get a sub array from elements that meet specific condition
-    public func subArray(_ condition: (Element) -> Bool) -> Array {
-        var subArray = Array()
-        for element in self {
-            if condition(element) {
-                subArray.append(element)
-            }
         }
         return subArray
     }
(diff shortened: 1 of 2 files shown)


Sources/ArrayExtensions.swift
```diff
@@ -19,6 +19,26 @@ public func ==<T: Equatable>(lhs: [T]?, rhs: [T]?) -> Bool {
 }
 
 extension Array {
+    
+    ///EZSE: Get a sub array from range of index
+    public func subArray(fromIndex:Int,toIndex:Int) -> Array {
```
piv199: Why this is better than `Array(array[fromIndex...toIndex])`

Sources/ArrayExtensions.swift
```diff
@@ -19,6 +19,26 @@ public func ==<T: Equatable>(lhs: [T]?, rhs: [T]?) -> Bool {
...
+        var subArray = Array()
+        for index in fromIndex...toIndex {
+            subArray.append(self[index])
+        }
+        return subArray
+    }
+    
+    ///EZSE: Get a sub array from elements that meet specific condition
+    public func subArray(_ condition: (Element) -> Bool) -> Array {
```
piv199: This is copy of `array.filter` function.

Sources/ArrayExtensions.swift
```diff
@@ -19,6 +19,15 @@ public func ==<T: Equatable>(lhs: [T]?, rhs: [T]?) -> Bool {
 }
 
 extension Array {
+    
+    ///EZSE: Get a sub array from range of index
+    public func get(atIndex:Int,toIndex:Int) -> Array {
```
piv199: think of `get(at range: ClosedRange<Int>) -> Array` 
And also you are missing checks for lower bound and upper bound
Thirdly you could implement it like `return self[range]`

piv199:
Also each new extension should have tests for it, we are currently trying to increase our code coverage.

Thanks for your pull request.
We are always appriciate any help.

lfarah:
@Dendim0n let us know if you need help writing the tests! 😉

piv199:
Could you please reset `Dendim0n:master` to `0ec8f56` commit?

Some refs:
[StackOverflow](http://stackoverflow.com/questions/17667023/git-how-to-reset-origin-master-to-a-commit)

piv199:
@lfarah could we approve it?

lfarah:
@piv199 I didn't like that ```testGet_SubArray``` with the underscore...

lfarah:
@piv199 sent you a message on Gitter about it

lfarah:
@Dendim0n thank you so much for helping out! 😃 👍

PR #262, opened 2016-10-20
<Description> : Adding random for CGFloat and tests for the same.

Khalian:
<Type> : feature/test

### DIFF
--- a/EZSwiftExtensionsTests/CGFloatTests.swift
+++ b/EZSwiftExtensionsTests/CGFloatTests.swift
@@ -8,6 +8,8 @@
 
 import XCTest
 
+@testable import EZSwiftExtensions
+
 class CGFloatTests: XCTestCase {
     var radians: CGFloat!
     var degrees: CGFloat!
@@ -45,5 +47,23 @@ class CGFloatTests: XCTestCase {
         mutableRadians = 2 * .pi
         mutableRadians.toDegreesInPlace()
         XCTAssertEqual(mutableRadians, degrees)
+    }
+
+    func testRandom() {
+        let randomCGFloat = CGFloat.random()
+        XCTAssertGreaterThanOrEqual(randomCGFloat, 0.0)
+        XCTAssertLessThanOrEqual(randomCGFloat, 1.0)
+    }
+    
+    func testRandomWithinRange() {
+        let range = CGFloat(0.0)...CGFloat(10.0)
+        let randomCGFloat = CGFloat.random(within: range)
+        XCTAssertGreaterThanOrEqual(randomCGFloat, 0.0)
+        XCTAssertLessThanOrEqual(randomCGFloat, 10.0)
+        
+        let closedRange = CGFloat(0.0)..<CGFloat(10.0)
+        let randomClosedCGFloat = CGFloat.random(within: closedRange)
+        XCTAssertGreaterThanOrEqual(randomClosedCGFloat, 0.0)
+        XCTAssertLessThanOrEqual(randomClosedCGFloat, 10.0)
     }
 }
--- a/Sources/CGFloatExtensions.swift
+++ b/Sources/CGFloatExtensions.swift
@@ -47,5 +47,19 @@ extension CGFloat {
     public static func radiansToDegrees(_ angleInDegrees: CGFloat) -> CGFloat {
         return (180.0 * angleInDegrees) / .pi
     }
-
+    
+    /// EZSE: Returns a random floating point number between 0.0 and 1.0, inclusive.
+    public static func random() -> CGFloat {
+        return CGFloat(Float(arc4random()) / 0xFFFFFFFF)
+    }
+        
+    /// EZSE: Returns a random floating point number in the range min...max, inclusive.
+    public static func random(within: Range<CGFloat>) -> CGFloat {
+        return CGFloat.random() * (within.upperBound - within.lowerBound) + within.lowerBound
+    }
+    
+    /// EZSE: Returns a random floating point number in the range min...max, inclusive.
+    public static func random(within: ClosedRange<CGFloat>) -> CGFloat {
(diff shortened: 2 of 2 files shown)


Sources/CGFloatExtensions.swift
```diff
@@ -47,5 +47,15 @@ extension CGFloat {
...
     }
-
+    
+    /// EZSE: Returns a random floating point number between 0.0 and 1.0, inclusive.
+    public static func random() -> CGFloat {
+        return CGFloat(Float(arc4random()) / 0xFFFFFFFF)
+    }
+    
+    /// EZSE: Returns a random floating point number in the range min...max, inclusive.
+    public static func random(min: CGFloat, max: CGFloat) -> CGFloat {
```
piv199: use ClosedRange or Range for this purposes. They ensure min < max.

Khalian:
I added the change. 

I have no idea why NSURL tests are failing. Perhaps create an issue to look into it?

lfarah:
@Khalian I think we need a better name for ```random()```. I wouldn't know that it could only be from 0 to 1 since floats can go bigger than just 1.

lfarah:
@piv199 what do you think?

lfarah:
@vsouza says that a good name would be ```randomFraction()```

Sources/CGFloatExtensions.swift
```diff
@@ -47,5 +47,14 @@ extension CGFloat {
     public static func radiansToDegrees(_ angleInDegrees: CGFloat) -> CGFloat {
         return (180.0 * angleInDegrees) / .pi
     }
-
+    
+    /// EZSE: Returns a random floating point number between 0.0 and 1.0, inclusive.
+    public static func random() -> CGFloat {
```
piv199: @lfarah suggests better name

As for me, I would stick with random because it's the default behavior.

Sources/CGFloatExtensions.swift
```diff
@@ -47,5 +47,14 @@ extension CGFloat {
...
+        return CGFloat(Float(arc4random()) / 0xFFFFFFFF)
+    }
+        
+    /// EZSE: Returns a random floating point number in the range min...max, inclusive.
+    public static func random(range: ClosedRange<CGFloat>) -> CGFloat {
```
piv199: random(in range:ClosedRange<CGFloat>) 
random(in range:Range<CGFloat>) 

what do you think?

lfarah: Hmmm I think these functions are good but random by itself is still useful.

piv199: @lfarah yes of course, I just point to 

`CGFloat.random(in: 300..<600.34)` instead of `CGFloat.random(range: 300..<600.34)`

And possibly add range if it not works, can't test. `300..<600.34` will work?

Khalian: So we have a consensus on this ? randomFraction with in: instead of range: ?

Khalian: I dont understand the previous comment.

piv199: @Khalian 0.123456789111315 vs 12345678911.1315

piv199: What actually the case for random to use. Where?

Khalian: Ok I can drop the random with range then.

Khalian: Ok so what do you want me to do at this point?

Khalian:
Closing this pull request as reviewers unable to reach a consensus.

lfarah:
@Khalian wanna open an issue and we can discuss and vote?

Sources/CGFloatExtensions.swift
piv199: `random(in range: ClosedRange<CGFloat>)`

EZSwiftExtensionsTests/CGFloatTests.swift
```diff
@@ -46,4 +48,17 @@ class CGFloatTests: XCTestCase {
...
     }
+
+    func testRandom() {
+        let randomCGFloat = CGFloat.random()
+        XCTAssertGreaterThanOrEqual(randomCGFloat, 0.0)
+        XCTAssertLessThanOrEqual(randomCGFloat, 1.0)
+    }
+    
+    func testRandomWithinRange() {
+        let range = CGFloat(0.0)...CGFloat(10.0)
```
piv199: Please, test also CGFloat(0.0)..<CGFloat(10.0)

Sources/CGFloatExtensions.swift
lfarah: @piv199 What if we deprecate this one and create a new? We already did this on some other methods:
``` swift
    /// EZSE: Checks if array contains at least 1 instance of the given object type
    @available(*, deprecated: 1.8, renamed: "containsType(of:)")
    public func containsInstanceOf<T>(_ element: T) -> Bool {
        return containsType(of: element)
    }

    /// EZSE: Gets the object at the specified index, if it exists.
    @available(*, deprecated: 1.8, renamed: "get(at:)")
    public func get(_ index: Int) -> Element? {
        return get(at: index)
    }

    /// EZSE: Checks if all elements in the array are true of false
    @available(*, deprecated: 1.8, renamed: "testAll(is:)")
    public func testIfAllIs(_ condition: Bool) -> Bool {
        return testAll(is: condition)
    }
```

EZSwiftExtensionsTests/CGFloatTests.swift
Khalian: Done. In is a keyword so used within.

piv199:
@Khalian Thanks very much for your patience 👍

lfarah:
Yes @Khalian, you're awesome! 🎉

PR #263, opened 2016-10-23
Add Tests to Date Extension

bcylin:
* Add more test cases to the existing date related methods
* Use a fixed reference to calculate time interval in between dates


lfarah:
Thanks @bcylin. Tests are always welcome! 😉

PR #264, opened 2016-10-23
Add Codecov Yaml

bcylin:
Ignore `EZSwiftExtensionsTests` to calculate code coverage based on the `Source` files only.

https://github.com/codecov/support/wiki/Codecov-Yaml
https://codecov.io/gh/goktugyil/EZSwiftExtensions

screenshot

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,4 @@
+coverage:
+  ignore:
+    - EZSwiftExtensionsExample/*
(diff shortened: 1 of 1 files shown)


lfarah:
@bcylin nice catch! 😉

goktugyil:
Oh man, much worse than we thought :(

lfarah:
@goktugyil that's for inspiration! We can do it 💪💪💪

PR #266, opened 2016-10-25
Changelog

piv199:
Initial changelog commit
Will do Unreleased part and could be merged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,80 @@ All notable changes to this project will be documented in this file.
 
 ### Added extensions
 
+- Array
+- `shuffled()` @piv199
+- `unique()` @Khalian
+- `get(at range:ClosedRange<Int>)` @Dendim0n
+
+- FloatingPoint
+- `round(toPlaces places: Int)` @piv199
+- `rounded(toPlaces places: Int)` @piv199
+- `ceil(toPlaces places: Int)` @piv199
+- `ceiled(toPlaces places: Int)` @piv199
+
+- String
+- `urlEncoded()` @Khalian
+- `urlEncode()` @Khalian
+
+- CGFloat
+- `radiansToDegrees()` @Khalian
+- `toDegreesInPlace()` @Khalian
+- `radiansToDegrees(_ angleInDegrees : CGFloat)` @Khalian
+- `random()` @Khalian
+- `random(within: Range<CGFloat>)` @Khalian
+- `random(within: ClosedRange<CGFloat>)` @Khalian
+
+
 ### Modified extensions
+
+- Updated `M_PI` to `.pi` #216 @piv199
+
+- Array
+- `reverseIndex(_ index: Int)` now returns `Int?` instead of `Int` @piv199
 
 ### Fixed bugs
 
+- Array
+- `shuffle()` crashing if array contains one or zero elements  @piv199
+
+- EZSwiftExtensions
+- Setting image on undefined thread for `imageWithURL` related methods #217 @piv199
+- `runThisEvery()` now returns `Timer` instead of `CFLoopTimer` #228 @lfarah
+
 ### Deprecated/Renamed extensions
 
+- Array
+- `removeObjects(_ objects: Element...)` renamed to `removeAll(_ elements: Element...)` @piv199
+- `removeFirstObject(_ object: Element)` renamed to `removeFirst(_ element: Element)` @piv199
+- `testIfAllIs(_ condition: Bool)` renamed to `testAll(is condition: Bool)` @piv199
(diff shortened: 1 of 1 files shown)


piv199:
@goktugyil @lfarah @Khalian 

Take a look at template [changelog.md](https://github.com/piv199/EZSwiftExtensions/blob/changelog/CHANGELOG.md)

- Should we change #### to just bold or even just text? Is it readable?
- Should we have a context menu for changelog section?
- Should we add tests change/adding/fixing messages to changelog?
- Should we add Danger/Codecov messages to changelog?
- Maybe change to table instead of list?
- Do we need to specify commit hash/url, author? issue related?

lfarah:
#### Should we change #### to just bold or even just text? Is it readable?
> Perfect the way it is right now
 
#### Should we have a context menu for changelog section?
> I didn't understand

#### Should we add tests change/adding/fixing messages to changelog?
I don't think so

#### Should we add Danger/Codecov messages to changelog?
> I don't think so

#### Maybe change to table instead of list?
> For me list is better

#### Do we need to specify commit hash/url, author? issue related?
> It's good to specify author, other things don't seem that important

piv199:
Updated, should we stick with such changelog?

lfarah:
@piv199 I think it's supposed to be all capital: ```CHANGELOG.md```. Besides that, it's perfect. Really good job! 🎉

piv199:
@lfarah it's already have that name

Khalian:
Should we change #### to just bold or even just text? Is it readable?

It looks all right. The purpose is legibility and #### amply satisfies that. 

Should we have a context menu for changelog section?

I think what you meant is what the change accomplishes (the purpose of a extension etc). In that case yes. Otherwise can you please elaborate?

Should we add tests change/adding/fixing messages to changelog?

No, and the reason I believe that is that the end user/programmers actually do not care about fixes and tests. 

Should we add Danger/Codecov messages to changelog?

No, similar reason as above. 

Maybe change to table instead of list?

The table is more concise. However if we wish to express context then a list is better way to do it. First we must arrive at the decision point whether we wish to include context. 

Do we need to specify commit hash/url, author? issue related?

The author to me is sufficient.

piv199:
Currently ** instead of ####

I ment a menu for section so we could easily jump to submenu "Added extensions" in 1.8. But your idea is interesting too...

I agree about tests, but not fixes in methods. If we fixed method we should add it.

I made renaming/deprecating with table because it's more concise, other is better in list variant.

commit hash maybe makes some sense, so when we look all changes we could easily jump to related pull request?

Khalian:
"I agree about tests, but not fixes in methods. If we fixed method we should add it."

My opinion would be if it modifies user experience then yes we should add. 

"I made renaming/deprecating with table because it's more concise, other is better in list variant."

Yes thats a good reason for using a table. 

"commit hash maybe makes some sense, so when we look all changes we could easily jump to related pull request"

I am not certain if end users will actually care about the commit hash. If they wanted to check the implementation they can directly jump into the source from XCode.

lfarah:
@piv199 can we push this?

piv199:
@lfarah I'l update and we will merge

goktugyil:
> Should we change #### to just bold or even just text? Is it readable?

I think the number would be better if it was bold

goktugyil:
> Should we change #### to just bold or even just text? Is it readable?

I think the number would be better if it was bold

> Should we have a context menu for changelog section?

No need. 

> Should we add tests change/adding/fixing messages to changelog?

Separate tests, irrelevant for the changelog. However something like "Increased test coverage by %4" might be good. 

> Should we add Danger/Codecov messages to changelog?

Messages? Like changes?

> Maybe change to table instead of list?


List

> Do we need to specify commit hash/url, author? issue related?

Lets give credit where its due. Even though github is good at it anyways.

piv199:
@goktugyil Okay, I'll add

piv199:
@lfarah I update as soon as I'm free, possibly tomorrow... so we could merge, and proceed with changelog in every pr.

lfarah:
@piv199 perfect!

Khalian:
Should we squash merge it at this time ?

piv199:
@Khalian yes, i forget how to do it properly

Khalian:
@piv199 rebase -i HEAD~23 on cmd and then replace the the picks with s for the 21 odd commits that you want to squash down. This will replay all the diffs into one big diff. While you do this it also will allow you to change the commit message should you choose to do so. Hope this helps.

lfarah:
@Khalian ?

Khalian:
It looks good minus the fact that there are 25 commits representing a Changelog as I type this.

piv199:
#344

PR #271, opened 2016-10-26
Renamed UIColor.random()

lfarah:
I changed ``` UIColor.randomColor()``` to ```UIColor.random()``` following the Swift3 guidelines.

#### Before Swift3:
```
UIColor.blackColor()
UIColor.greenColor()
```

#### After Swift3:
```
UIColor.black
UIColor.green
```

### DIFF
--- a/Sources/UIColorExtensions.swift
+++ b/Sources/UIColorExtensions.swift
@@ -57,7 +57,6 @@ extension UIColor {
     public var alpha: CGFloat {
         var a: CGFloat = 0
         getRed(nil, green: nil, blue: nil, alpha: &a)
-        
         return a
     }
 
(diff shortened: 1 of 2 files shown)


Sources/UIColorExtensions.swift
```diff
@@ -61,7 +61,7 @@ extension UIColor {
     }
 
     /// EZSE: Returns random UIColor with random alpha(default: false)
-    public static func randomColor(_ randomAlpha: Bool = false) -> UIColor {
+    public static func random(_ randomAlpha: Bool = false) -> UIColor {
```
piv199: Imagine `UIColor.random(true)` in code, it's not obvious as for me

lfarah:
@piv199 now we have:
``` swift
        let color = UIColor.random(randomAlpha: true)
```

PR #272, opened 2016-10-28
<Description> : Implementing CGFloat shortest angle distance in radians.

Khalian:
<Type> : feature.

### DIFF
--- a/Sources/CGFloatExtensions.swift
+++ b/Sources/CGFloatExtensions.swift
@@ -62,4 +62,22 @@ extension CGFloat {
     public static func random(within: ClosedRange<CGFloat>) -> CGFloat {
         return CGFloat.random() * (within.upperBound - within.lowerBound) + within.lowerBound
     }
+    
+    /**
+      EZSE :Returns the shortest angle between two angles. The result is always between
+      -π and π.
+     
+      Inspired from : https://github.com/raywenderlich/SKTUtils/blob/master/SKTUtils/CGFloat%2BExtensions.swift
+     */
+    public static func shortestAngleInRadians(from first: CGFloat, to second: CGFloat) -> CGFloat {
+        let twoPi = CGFloat(.pi * 2.0)
+        var angle = (second - first).truncatingRemainder(dividingBy: twoPi)
+        if (angle >= .pi) {
+            angle = angle - twoPi
+        }
+        if (angle <= -.pi) {
+            angle = angle + twoPi
+        }
+        return angle
+    }
 }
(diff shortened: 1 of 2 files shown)


Sources/CGFloatExtensions.swift
```diff
@@ -62,4 +62,22 @@ extension CGFloat {
...
         return CGFloat.random() * (within.upperBound - within.lowerBound) + within.lowerBound
     }
+    
+    /**
+      EZSE :Returns the shortest angle between two angles. The result is always between
+      -π and π.
+     
+      Inspired from : https://github.com/raywenderlich/SKTUtils/blob/master/SKTUtils/CGFloat%2BExtensions.swift
+     */
+    public static func shortestAngleBetweenInRadians(angle1InRadians: CGFloat, angle2InRadians: CGFloat) -> CGFloat {
```
piv199: I feel strange of such naming...
Does this extension is helpful?

Khalian: So the is the question about whether this extension adds any value to this project or can we provide it a better name?

piv199: both

piv199: CGFloat.shortestAngleInRadians(_ angle1: firstAngle, _ angle2: fullCircleAngle)
CGFloat.shortestAngleInRadians(from angle1: firstAngle, to angle2: fullCircleAngle)
?

Khalian: Answer to the second question : 

If CGFloats are treated as angles in the Core Graphics package(which our existing extensions like degreesToRadians actually do) , there should exist at least one method that computes the angular distance between the two.

Answer to the first question : -

The first option seems like a valid alternative.

So are we in a situation to reach consensus?

Sources/CGFloatExtensions.swift
```diff
@@ -62,4 +62,22 @@ extension CGFloat {
...
         return CGFloat.random() * (within.upperBound - within.lowerBound) + within.lowerBound
     }
+    
+    /**
+      EZSE :Returns the shortest angle between two angles. The result is always between
+      -π and π.
+     
+      Inspired from : https://github.com/raywenderlich/SKTUtils/blob/master/SKTUtils/CGFloat%2BExtensions.swift
+     */
+    public static func shortestAngleInRadians(fromAngle: CGFloat, toAngle: CGFloat) -> CGFloat {
```
piv199: shortestAngleInRadians(from angle: CGFloat, to angle: CGFloat) -> CGFloat

Khalian: Thats a syntax error : Definition conflicts with previous values.

piv199: @Khalian what do you mean?

Khalian: Swift does not allow spaces in the names of the variables involved. The definition conflicts come because from Angle and to Angle define Angle twice. Thats why I used camel case.

piv199: shortestAngleInRadians(from firstAngle: CGFloat, to secondAngle: CGFloat) -> CGFloat

Khalian: Ah ok. Done.

piv199:
@Khalian actually you are the best, your help matters to this repo. Thanks!

PR #273, opened 2016-10-28
Added more String tests

lfarah:
More tests = 🎉👻😎✅🤑😺

### DIFF
--- a/EZSwiftExtensionsTests/StringTests.swift
+++ b/EZSwiftExtensionsTests/StringTests.swift
@@ -253,4 +253,55 @@ class StringTests: XCTestCase {
         
         XCTAssertEqual("a%20b%20c%20d%20e", spacedString)
     }
+    
+    func testSplitSeparator() {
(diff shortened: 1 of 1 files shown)


EZSwiftExtensionsTests/StringTests.swift
```diff
@@ -253,4 +253,55 @@ class StringTests: XCTestCase {
...
+        
+        let testString = "H3yH0L3tsG0"
+        let expectedResult = ["3","0","3","0"]
+        
+        XCTAssertEqual(testString.matchesForRegexInText("[0-9]"), expectedResult)
+    }
+    
+    func testIsNumber() {
+        
+        let testString1 = "E"
```
Khalian: Can we add a number which is a float/double e.g 3.14 or is that already covered?

lfarah: @Khalian done✅

PR #278, opened 2016-10-30
<Description> : Adding CGPoint Tests : For adding two CGPoints and calculating the distance between the two.

Khalian:
<Type> : feature/test

### DIFF
--- a/Sources/CGPointExtensions.swift
+++ b/Sources/CGPointExtensions.swift
@@ -1,0 +1,24 @@
+//
+//  CGPointExtensions.swift
+//  EZSwiftExtensions
+//
+//  Created by Sanyal, Arunav on 10/29/16.
+//  Copyright © 2016 Goktug Yilmaz. All rights reserved.
+//
+
+import UIKit
+import Foundation
+
+extension CGPoint {
+    
+    /// EZSE: Converts angle degrees to radians.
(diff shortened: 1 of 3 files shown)


Sources/CGPointExtensions.swift
```diff
@@ -0,0 +1,24 @@
...
+extension CGPoint {
+    
+    /// EZSE: Converts angle degrees to radians.
+    public static func + (this: CGPoint, that: CGPoint) -> CGPoint {
+        return CGPoint(x: this.x + that.x, y: this.y + that.y)
+    }
+    
+    /// EZSE: Calculates the distance between two CG Points. 
+    /// Inspired by : http://stackoverflow.com/questions/1906511/how-to-find-the-distance-between-two-cg-points
+    public static func absDistance(this: CGPoint, that: CGPoint) -> CGFloat {
```
piv199: why call it absDistance if distance is already abs'ed?

Khalian: should i just name it distance?

piv199: @Khalian I'm for it

piv199: also this/that is not appropriative - use from/to or start/end

PR #279, opened 2016-10-31
<Description> : Adding normalization of the vector representation of CGPoint.

Khalian:
<Type> : feature/test


goktugyil:
Thanks 👍 

Would you mind adding more tests, both to this and the distance one. Like maybe 5:12:13 and others.

PR #280, opened 2016-10-31
Added more IntTests

lfarah:
Tests added:
* ```testDigits()``` (more coverage)
*  ```func testRange()``` (new)
* ``` func testString()``` (new)

### DIFF
--- a/EZSwiftExtensionsTests/IntTests.swift
+++ b/EZSwiftExtensionsTests/IntTests.swift
@@ -17,7 +17,14 @@ class IntTests: XCTestCase {
     }
 
     func testDigits() {
-        XCTAssertEqual(value.digits, 5)
+        
+        let value1: Int = 0
+        let expected1: Int = 1
+        let value2: Int = 50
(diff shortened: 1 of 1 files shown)


EZSwiftExtensionsTests/IntTests.swift
```diff
@@ -79,4 +86,26 @@ class IntTests: XCTestCase {
...
+        
+        XCTAssertEqual(value1.toString, expected1)
+        XCTAssertEqual(value2.toString, expected2)
+        XCTAssertEqual(value3.toString, expected3)
+    }
+    
+    func testRange() {
+        
+        let value: Int = 0
+        let expected: CountableRange<Int> = 0..<value
```
piv199: do not rely on value, create expected range with constants.

Also we need some at least test where range is valid :)

lfarah: True, totally messed up. Sending an update right now

PR #283, opened 2016-10-31
Added testAppending to NSAttributedStringTests

lfarah:
```testAppending()``` tests ```NSAttributedString() += NSAttributedString()```


goktugyil:
Also, as an addition you might want to add a case where strings with different attributes are added together

PR #290, opened 2016-11-04
<Description> : Adding the dot product operation for the vector representation of CGPoints.

Khalian:
## Checklist

- [x] New Extension
- [x] New Test


goktugyil:
I think we shouldn't a new operator for this. After all we're not a graphics library and if a person wants to do this with their custom operator, they easily can. I think the name should be intuitive and readable like all functions.

lfarah:
@goktugyil can we close this?

Khalian:
In that case should I refactor this to func dotProd instead of the operator?

lfarah:
@piv199? @goktugyil?

goktugyil:
@Khalian  yes like this

PR #298, opened 2016-11-05
Added testFrame() and testFontSize()

lfarah:
## Checklist

- [ ] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/UITextFieldTests.swift
+++ b/EZSwiftExtensionsTests/UITextFieldTests.swift
@@ -7,10 +7,23 @@
 //
 
 import XCTest
+@testable import EZSwiftExtensions
 
 class UITextFieldTests: XCTestCase {
 
-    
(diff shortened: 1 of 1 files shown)


EZSwiftExtensionsTests/UITextFieldTests.swift
```diff
@@ -7,10 +7,23 @@
...
 
 import XCTest
+@testable import EZSwiftExtensions
 
 class UITextFieldTests: XCTestCase {
 
-    
+    func testFrame() {
+        
+        let textField = UITextField(x: 30, y: 35, w: 40, h: 45)
```
Khalian: Minor : We can reuse the values 30, 35 etc by declaring them exactly once.

This way if anyone wants to extend tests and use them, they wont have to write it again.

lfarah:
@Khalian done! ✅

PR #302, opened 2016-11-07
Uilabel changes

vilapuigvila:
## Checklist

- [ ] New Extension
- [ ] New Test

### DIFF
--- a/EZSwiftExtensionsTests/UILabelTests.swift
+++ b/EZSwiftExtensionsTests/UILabelTests.swift
@@ -23,15 +23,15 @@ class UILabelTests: XCTestCase {
     func testText() {
         
         let label = UILabel(frame: CGRect(x: 0, y: 0, width: 200, height: 50))
-        label.setText("EZSwiftExtensions✅", animated: false, duration: nil)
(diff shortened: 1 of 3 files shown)


lfarah:
Hey @vilapuigvila, can you please explain your changes?

vilapuigvila:
Glups... I'm sorry. Maybe I should not have done that without first consulting. Anyway my idea was to remove animate param bool, becasuse I think is unnecessary if you give a 0 value in duration param. This way we saved to enter a parameter, we can write this label.animateText("text", duration: 0) instead label.setText("", animated: false, duration: 0)

lfarah:
@vilapuigvila makes total sense on the boolean.
Before approving the PR, I just need a couple things from you:
* Put the ```cachesDirectoryPath()``` and ```documentsDirectoryPath()``` in another PR
* Change the method name to ```label.set(text: "text", duration: 0)``` so we're following Swift 3 naming
* Update ```UILabelTests.swift``` with the new method name

piv199:
Please reverse hunk in UIDevice extension

Khalian:
Closing this CR and opening this instead : https://github.com/goktugyil/EZSwiftExtensions/pull/340

PR #305, opened 2016-11-08
<Description> : Adding day, month and year extensions to date.

Khalian:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/CGPointTests.swift
+++ b/EZSwiftExtensionsTests/CGPointTests.swift
@@ -61,4 +61,15 @@ class CGPointTests: XCTestCase {
         let onNegativeXAxisP = CGPoint(x:-1, y:0)
         XCTAssertEqual(onNegativeXAxisP.angle, .pi)
     }
+    
+    func testDotProduct() {
+        let p = CGPoint(x:1, y:2)
+        let q = CGPoint(x:3, y:4)
+        
+        XCTAssertEqual(CGPoint.dotProduct(this: p, that: q), 11)
+        
+        let zero = CGPoint(x:0, y:0)
+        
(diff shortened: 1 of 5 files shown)


lfarah:
@Khalian do you wanna also add the **month*  as String? Also index + String of the current day of the week would be good.
``` swift
    // EZSE : Get the weekday from the date
     public var week: String {
		let dateFormatter = NSDateFormatter()
		dateFormatter.dateFormat = "EEEE"
		return dateFormatter.stringFromDate(self)
    }

**Comment by Khalian** [comment]
Ok, I can put the week in. 

I dont understand why the month should be string? The gregorian calendar has different syntax across different cultures. An integer makes more sense as it disambiguates.

**Comment by goktugyil** [comment]
I think month makes more sense as int too. But there might be some use cases for string, idk.

**Comment by lfarah** [comment]
I think we should have both **Int** and **String** for *month* and *day of the week*

**Comment by goktugyil** [comment]
Wouldn't hurt I guess

**Comment by Khalian** [comment]
No I dont understand how I am going to implement a string based month here. 

Will it be locale specific (e.g. for spanish have a struct with all of the months and return it accordingly), similarly for english and so on.

**Comment by lfarah** [comment]
@Khalian ```NSDateFormatter``` gives us the month name in the right locale:
``` swift
extension NSDate {
    var month: String {
        let dateFormatter = NSDateFormatter()
        dateFormatter.dateFormat = "MMMM"
        return dateFormatter.stringFromDate(self)
    }
}
```

Khalian:
Added weekday and monthAsString.

lfarah:
Thanks @Khalian! 👍 
Can you please remove the changes on ```CGPointExtensions.swift```,```UITextFieldTests.swift``` and ```CGPointTests.swift```?

PR #312, opened 2016-11-10
Uidevice add extensions

vilapuigvila:
## Checklist

- [ ] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/UIDeviceTests.swift
+++ b/EZSwiftExtensionsTests/UIDeviceTests.swift
@@ -7,10 +7,29 @@
 //
 
 import XCTest
+@testable import EZSwiftExtensions
 
 class UIDeviceTests: XCTestCase {
 
+    override func setUp() {
+        super.setUp()
+    }
     
-
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    func testGetDocumentsDirectoryPath() {
+        
+        let docPath = UIDevice.documentsDirectoryPath()
+        XCTAssert(docPath.characters.count > 1)
+    }
+    
+    func testGetCachesDirectoryPath() {
+        
+        let docPath = UIDevice.cachesDirectoryPath()
+        XCTAssert(docPath.characters.count > 1)
+    }
+    
 }
 
--- a/EZSwiftExtensionsTests/UIDeviceTests.swift
+++ b/EZSwiftExtensionsTests/UIDeviceTests.swift
@@ -21,14 +21,24 @@ class UIDeviceTests: XCTestCase {
     
     func testGetDocumentsDirectoryPath() {
         
-        let docPath = UIDevice.documentsDirectoryPath()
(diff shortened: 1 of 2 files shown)


Sources/UIDeviceExtensions.swift
```diff
@@ -85,6 +85,18 @@ extension UIDevice {
         return UIDevice.current.userInterfaceIdiom == UIUserInterfaceIdiom.pad
     }
 
+    /// EZSE: Returns path of documents directory
+    class func documentsDirectoryPath() -> String {
```
Khalian: public class func.

Sources/UIDeviceExtensions.swift
```diff
@@ -85,6 +85,18 @@ extension UIDevice {
...
+        let paths = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true) as [String]
+        return paths[0]
+    }
+    
+    /// EZSE: Returns path of documents directory caches
+    class func cachesDirectoryPath() -> String {
```
Khalian: public class func

Sources/UIDeviceExtensions.swift
Khalian: I dont understand how this is a UIDeviceExtension? There is no reference to self or any static references to UIDevice anywhere here.

vilapuigvila: @Khalian I agree, where do you thing would be the best place to put it? Maybe NSFileManager?

Khalian: Yea that sounds about right. I would wait for one of the collaborators to comment.

EZSwiftExtensionsTests/UIDeviceTests.swift
```diff
@@ -7,10 +7,29 @@
...
+        super.setUp()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    func testGetDocumentsDirectoryPath() {
+        
+        let docPath = UIDevice.documentsDirectoryPath()
```
Khalian: This test does not really convey what exactly documentsDirectoryPath is supposed to accomplish.

Sources/UIDeviceExtensions.swift
lfarah: I agree. Should be FileManagerExtensions.

@goktugyil? @piv199?

goktugyil: Ok, you need me to create the file?

Khalian: No I dont think its needed. I added CGPoint extensions before. So this project does not stop anyone.

goktugyil: There is a little trick to make sure new files are referenced properly in normal project, example project and tests. People usually miss some part, I should write a checklist about it sometime

vilapuigvila:
This pull request can be closed, I moved code to a new branch.

PR #327, opened 2016-11-27
Class fun to print family fonts and font names

vilapuigvila:
## Checklist

- [ ] New Extension
- [ ] New Test

### DIFF
--- a/Sources/UIFontExtensions.swift
+++ b/Sources/UIFontExtensions.swift
@@ -78,6 +78,16 @@ extension UIFont {
       return nil
     }
 
+    /// EZSwiftExtensions print all family fonts and font names
+    public class func PrintAllFonts() {
+        let fontFamilyNames = self.familyNames
+        for familyName in fontFamilyNames {
+            print("Font Family Name = [\(familyName)]")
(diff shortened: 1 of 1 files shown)


piv199:
I can't see a good use case for this method, because when it's printed it doesn't reflect the style. And if you would like to know all family fonts you could refer to http://iosfonts.com .

vilapuigvila:
@piv199 I used to print fonts when I added extra fonts in xcode, sometimes font name doesn't match with the print name, and then I waste time to get font.. I didn't know http://iosfonts.com , great website!!

Khalian:
I am more bothered about how to test this. Can you confirm that this print method does has have different output across swift versions or devices?

lfarah:
We already deprecated [that function](https://github.com/goktugyil/EZSwiftExtensions/blob/master/Sources/UIFontExtensions.swift#L97)

Khalian:
Hey would it make sense to include static methods for all the font families (getters using CGFloat)? Or is it overkill ?

lfarah:
@Khalian I think it's overkill

vilapuigvila:
right, so you can close this PR. Thanks

PR #331, opened 2016-12-03
<Description> : Adding uppercase and lower case for character extensions.

Khalian:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/CharacterTests.swift
+++ b/EZSwiftExtensionsTests/CharacterTests.swift
@@ -13,6 +13,7 @@ import XCTest
 class CharacterTests: XCTestCase {
     
     let pchr: Character = "p"
+    let pchrUpper: Character = "P"
     let digitChr: Character = "3"
     
     let int: Int = 3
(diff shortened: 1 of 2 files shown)


lfarah:
I think it's better if we don't camel case ```lowerCase``` and ```upperCase``` since Apple already uses it lowercased (no puns intended 😂): ```lowercased()``` and ```uppercased()```

Khalian:
Done.

EZSwiftExtensionsTests/CharacterTests.swift
```diff
@@ -30,4 +31,14 @@ class CharacterTests: XCTestCase {
         XCTAssertEqual(digitChr.toInt, int)
         XCTAssertEqual(pchr.toInt, nil)
     }
+    
+    func testUpperCase() {
```
lfarah: Should change name for tests too

EZSwiftExtensionsTests/CharacterTests.swift
```diff
@@ -30,4 +31,14 @@ class CharacterTests: XCTestCase {
...
+        XCTAssertEqual(pchr.uppercased, pchrUpper)
+        XCTAssertEqual(digitChr.uppercased, digitChr)
+    }
+    
+    func testLowerCase() {
```
lfarah: Same here

Khalian: Done.

piv199:
`String().lowercased()` but `Character().lowercased`, why?
Also, doesn't String and Character have common protocol to extend with this function?

PR #332, opened 2016-12-04
Added new extension "navigationBarHidden".

vilapuigvila:
## Checklist

- [x] New Extension
- [x] New Test


lfarah:
I liked it because ```setNavigationBarHidden``` is not Swifty, so this extension is useful!

goktugyil:
navigationBarHidden is also not very clear cause its requiring a variable after that. I expect this function to return true or false depending on if nav bar is hidden. 

How about, 
hideNavigationBar() 
showNavigationBar()

navigationBarVisibility(_hide: Bool)

?

Also you should get a swiftlint error from 'hide:Bool'

vilapuigvila:
We can do isNavigationBarHidden? like "isHidden", "isUserInteractionEnabled".
What do you think?

goktugyil:
Yeah that might work

Khalian:
@vilapuigvila Would you mind merging these changes and adding a CHANGELOG entry. I will merge it after those steps.

Khalian:
I am performing a manual merge on this (via command line) and maybe use another PR. I will also add a CHANGELOG entry. I do not presume there would be an issue since its approved by two users.

Khalian:
https://github.com/goktugyil/EZSwiftExtensions/pull/428

PR #343, opened 2016-12-15
<Description> : Adding is in future and is in past extensions.

Khalian:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/Sources/DateExtensions.swift
+++ b/Sources/DateExtensions.swift
@@ -181,21 +181,26 @@ extension Date {
 
     // EZSE : Get the day from the date
     public var day: Int {
-        return NSCalendar.current.component(Calendar.Component.day, from: self)
+        return Calendar.current.component(.day, from: self)
     }
 
     /// EZSE: Get the hours from date
     public var hour: Int {
-        return NSCalendar.current.component(Calendar.Component.hour, from: self)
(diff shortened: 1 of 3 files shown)


Sources/DateExtensions.swift
```diff
@@ -111,6 +111,16 @@ extension Date {
             return "\(components.second) seconds ago"
         }
     }
+    
+    /// EZSE: Check if date is in future.
+    public var isInFuture: Bool {
```
piv199: isFuture

Sources/DateExtensions.swift
```diff
@@ -111,6 +111,16 @@ extension Date {
...
+        return self > Date()
+    }
+    
+    /// EZSE: Check if date is in past.
+    public var isInPast: Bool {
```
piv199: isPast

Sources/DateExtensions.swift
Khalian: Done

goktugyil:
In what cases would we need this?

Khalian:
Off the top of my head running cron jobs or a trigger after a particular date.

piv199:
Could you also add to changelog, but maybe not in this pr

Khalian:
We can punt this to the next release and tag the last release as official.

PR #347, opened 2016-12-24
<Description> : Adding lcm and gcd for unsigned integer.

Khalian:
<Type> : feature/test

## Checklist

- [x] New Extension
- [x] New Test


piv199:
Add entry to changelog please, I forgot to merge rule from danger branch into master

Khalian:
Cool. I will add it after we pull in https://github.com/goktugyil/EZSwiftExtensions/pull/348. I wont merge till then.

piv199:
Hm, I thought about that we do not actually need labels for each parameter, because they don't give any sense.

Also, what's the best way to find gcd of three or more elements? Like, `Int.gcd(x1,x2,x3...)`?
What do you think?

Khalian:
gcd(a,b,c)=gcd(a,(gcd(b,c)) but if I have to introduce a var arg way of getting gcd, I will introduce it as a separate method. 

Euclid's algo solves only for 2 numbers. I don't know any algorithm on the top of my head with better computational complexity than it.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 
 Add subsequent changes that you make in this section. 
 
+2. **UInt**
+ - `static gcd(firstNum:UInt, secondNum:UInt) -> UInt` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/347) by *Khalian*
```
piv199: Add `_` before labels

Khalian:
Done.

CHANGELOG.md
Khalian: Done.

Sources/IntExtensions.swift
```diff
@@ -51,10 +51,25 @@ extension Int {
...
-
 }
 
 extension UInt {
     /// EZSE: Convert UInt to Int
     public var toInt: Int { return Int(self) }
+    
+    /// EZSE: Greatest common divisor of two integers using the Euclid's algorithm.
+    /// Time complexity of this in O(log(n))
+    public static func gcd(_ firstNum:UInt, secondNum:UInt) -> UInt {
```
piv199: Still do not see the purpose of secondNum label, when we autocomplete this function we have two parameters of type UInt. We could infer that the first if firtstUint, and second is secondUint :)

Sources/IntExtensions.swift
```diff
@@ -51,10 +51,25 @@ extension Int {
...
+        let remainder = firstNum % secondNum
+        if remainder != 0 {
+            return gcd(secondNum, secondNum: remainder)
+        } else {
+            return secondNum
+        }
+    }
+    
+    /// EZSE: Least common multiple of two numbers. LCM = n * m / gcd(n, m)
+    public static func lcm(_ firstNum:UInt, secondNum:UInt) -> UInt {
```
piv199: The same

Khalian: Done

Sources/IntExtensions.swift
Khalian: Done.

PR #349, opened 2016-12-24
<Description> : Adding vector and angle based constructors for CGPoint.

Khalian:
<Type> : feature/test

## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 
 Add subsequent changes that you make in this section. 
 
+1. **CGPoint:**
+   - `init(vector: CGVector)` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/349) by *Khalian*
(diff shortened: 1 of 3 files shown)


piv199:
Does it make any sense? I'm not used to this functions. I can't approve/disapprove this pr :)

Khalian:
@piv199 huh ?

piv199:
How could we get point from angle? Is it logically normal?)

Khalian:
Ah yes, its normal. 

self.init(x: cos(angle), y: sin(angle) will create the point. If you take the length from origin = x^2 + y^2 we get Cos^2 (angle) + Sin^2(angle) which is always 1. 

This extension is effectively a polar coordinate to cartesian transform minus the scaling factor (to keep it simple).

Khalian:
Actually I don;t really mind it being a proper coordinate transform function (that would have a CGFloat scaling factor), but its upto you guys if thats a useful extension?

PR #352, opened 2017-01-01
<Description> : Add toggled var to Bool.

Khalian:
<Type> : feature/test

## Checklist

- [x] New Extension
- [x] New Test


EZSwiftExtensionsTests/BoolTests.swift
```diff
@@ -22,5 +22,13 @@ class BoolTests: XCTestCase {
         value.toggle()
         XCTAssertNotEqual(value, false)
     }
+    
+    func testToggled() {
+        let value = false
```
lfarah: if the other variable is trueVal, this one should be falseVal to keep consistency

Khalian: Done.

PR #354, opened 2017-01-03
vilapuigvila array

vilapuigvila:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -247,4 +247,14 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(duplicatedElementsArrayOrderMaintained.unique(), [2, 1, 3])
         
     }
+    
+    func testRemoveEqualItems() {
+        var list = ["one", "two", "two", "four"]
+        list.removeEqualItems(item: "two")
+        
+        XCTAssertEqual(["one", "four"], list)
+        
+        list.removeEqualItems(item: "five")
+        XCTAssertEqual(["one", "four"], list)
+    }
 }
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -250,11 +250,11 @@ class ArrayTests: XCTestCase {
     
     func testRemoveEqualItems() {
         var list = ["one", "two", "two", "four"]
-        list.removeEqualItems(item: "two")
+        list.removeAllItems(with: "two")
         
         XCTAssertEqual(["one", "four"], list)
         
-        list.removeEqualItems(item: "five")
+        list.removeAllItems(with: "five")
         XCTAssertEqual(["one", "four"], list)
     }
 }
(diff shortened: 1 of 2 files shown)


EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -247,4 +247,14 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(duplicatedElementsArrayOrderMaintained.unique(), [2, 1, 3])
         
     }
+    
+    func testRemoveEqualItems() {
+        var list = ["one", "two", "two", "four"]
+        list.removeEqualItems(item: "two")
+        
+        XCTAssertEqual(["one", "four"], list)
```
Khalian: My first thought was that this API will remove duplicates of an item. It seems that is not true.

Could you refactor the name of this method to removeItemsWithValue?

Sources/ArrayExtensions.swift
```diff
@@ -132,6 +132,13 @@ extension Array where Element: Equatable {
         self.remove(at: index)
     }
 
+    /// EZSE: Removes equal items of array
```
Khalian: Change the comment to removes all items in an array with a particular value.

EZSwiftExtensionsTests/ArrayTests.swift
lfarah: @Khalian what do you think about making it more Swift 3?
``` swift
func removeEqualItems(with value: Element) {
    
}
```

Khalian: @lfarah 

   func removeAllItems(with value:Element) {

   }

is more appropriate

piv199:
This is unnecessary extension, thus in swift we have: `array.filter { $0 != "two" }`, `array.filter { !["two", "three"].contains($0) }`

Khalian:
@piv199 yea, most people do not understand anonymous methods though. 

Actually now I am kind of on the fence. @lfarah do you think this method adds value considering filters are trivial to implement in swift ?

vilapuigvila:
@piv199 I'm agree, but if we use filter we need store filtered data in a new array, the original array still the same objects. My idea was to modify and work with one array. 
let newArray = list.filter({
    $0 != "two"
 })

Sources/ArrayExtensions.swift
vilapuigvila: Ok, I'm going to do that. Should be a good idea to do your first thought?
 func removeRepeatedElemets()

Khalian:
Cool. Now would you mind squashing these commits and adding an entry to the CHANGELOG.md for this?

furuyan:
Could you tell me the difference with `func removeAll(_ elements: Element...)`?
It looks equal function.

lfarah:
I do think this method adds value

EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -247,4 +247,14 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(duplicatedElementsArrayOrderMaintained.unique(), [2, 1, 3])
         
     }
+    
+    func testRemoveEqualItems() {
```
Khalian: Refactor name to testRemoveAllItemsWithValue.

vilapuigvila:
@furuyan 👍  you're right is the same, so this PR can be closed, I'm sorry..

PR #376, opened 2017-02-06
support tvOS

KenShih522:
## Checklist

- [x] New Extension
- [ ] New Test

### DIFF
--- a/Sources/StringExtensions.swift
+++ b/Sources/StringExtensions.swift
@@ -450,11 +450,16 @@ extension String {
         return false
     }
     
+    /// UIPasteboard is __TVOS_PROHIBITED and hence only supported in iOS
+    #if os(iOS)
+    
     /// EZSE: copy string to pasteboard
     func addToPasteboard() {
         let pasteboard = UIPasteboard.generalPasteboard()
         pasteboard.string = self
     }
+    
+    #endif
 }
 
 /// EZSE: Pattern matching of strings via defined functions


KenShih522:
support tvOS

Khalian:
I am confused as to what this extension does. You want to disable or enable tvOS support for addToPasteboard. If so why?

KenShih522:
Hi, i am developing an application for tvOS. I found that I can not compile your source code on tvOS.
It seems like UIPasteboard can not be used on tvOS because they marked it __TVOS_PROHIBITED.

Sources/StringExtensions.swift
```diff
@@ -450,11 +450,15 @@ extension String {
         return false
     }
     
+    #if os(iOS)
```
Khalian: Lets add a comment on top of this as to why this change is being made ("UIPasteboard is __TVOS_PROHIBITED and hence only supported in iOS)"

Khalian:
Also would you mind changing the description of the commit message to something along the lines of:-

"Defining addToPasteboard StringExtension to only be usable in IOS

Reason : UIPasteboard is __TVOS_PROHIBITED"

Khalian:
Before we merge, @lfarah I think swift lint is acting up again. Do you have any context on this ?

lfarah:
@Khalian newest Xcode doesn't support Swift 2.3

lfarah:
@KenShih522 do you intend to use Swift 2.3 or that was a mistake?

KenShih522:
Yes, I am currently using swift 2.3 to develop my project.

Khalian:
@goktugyil should we merge this ?

goktugyil:
tvOS support constantly fails while we add new things. IS there like an auto test that checks for that too?

Khalian:
@goktugyil We can raise autotests for tvos as a separate issue. Resolving this for now.

PR #375, opened 2017-02-05
<Description> : Adding ISO8601 Date formatted var extension for Date.

Khalian:
<Type> : feature/test

## Checklist

- [x] New Extension
- [x] New Test
- [ ] Changed more than one extension, but all changes are related
- [ ] Trivial change (doesn't require changelog)

PR #378, opened 2017-02-09
Vilapuigvila arrayextension

vilapuigvila:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -247,4 +247,19 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(duplicatedElementsArrayOrderMaintained.unique(), [2, 1, 3])
         
     }
+    
+    func testElementForIndexPath() {
+        let array = [["a", "b", "c"],["1", "2"]]
+        let item  = array.elemetForIndexPath(section: 0, row: 0) as? String
+	    XCTAssertEqual(item, "a")
+        
+        let noSection = array.elemetForIndexPath(section: 2, row: 0) as? String
+        XCTAssertNil(noSection)
+        
+        let noRow = array.elemetForIndexPath(section: 1, row: 5) as? String
+        XCTAssertNil(noRow)
+        
+        let secondRow = array.elemetForIndexPath(section: 1, row: 1) as? String
+        XCTAssertEqual(secondRow, "2")
+    }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,19 +18,25 @@ Add subsequent changes that you make in this section.
    - `var isFuture` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/343) by *Khalian*
    - `var isPast` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/343) by *Khalian*
 
-4. **Float**
+4. **Double**
+   - `var abs` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/365) by *Khalian*
+
+5. **Float**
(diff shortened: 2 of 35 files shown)


Khalian:
Please rebase mainline to your branch and resend the PR. I am having a very hard time reading this.

vilapuigvila:
@Khalian done.

Khalian:
Its still not clean enough. Please squash your commits, replay in on top by mainline by rebasing it so that the PR appears as exactly one commit. Then resend it. The expectation at our end is we see only your changes and nothing else.

EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -247,4 +247,19 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(duplicatedElementsArrayOrderMaintained.unique(), [2, 1, 3])
         
     }
+    
+    func testElementForIndexPath() {
+        let array = [["a", "b", "c"],["1", "2"]]
```
Khalian: So what does this array represent? An M * N matrix (in your test case its 3 * 2). If so is it a matrix, and I presume section is the row number and column number, how is it represented in this array. 

Traditionally you represent it via a either a row major or a column major format https://en.wikipedia.org/wiki/Row-_and_column-major_order. Your representation is neither of those. 

The other possibility is its a cartesian product. In which case this test case is wrong.

In either case I would not vouch for this extension. From a Data structure standpoint, its better design to present a matrix struct with checks and balances to prevent data from being misrepresented (for instance utilizing this method on a 3 dimensional matrix would not make sense).

vilapuigvila:
@Khalian  I've squashed the last four commits into one, I don't have too experience in interactice rebase mode.

vilapuigvila:
@Khalian I used this extension in tableview or collectionview datasource when I they have several sections and my data in one multidimensional array, for instance:
 let item = items.elemetForIndexPath(section: indexPath.section, row: indexPath.row)
cell.loadData(item)

Khalian:
"I don't have too experience in interactice rebase mode."

As a contributor, its your responsibility to make sure that the PR that you submit can actually be pulled and merged. Otherwise the contract you set up with reviewers is that we have to do the legwork for cleaning up and making the commit useable in the context of a merge post review.  

Short of giving you references on different git workflows, I honestly do not know how I can make this any more clearer.  

https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

Khalian:
"let item = items.elemetForIndexPath(section: indexPath.section, row: indexPath.row)
cell.loadData(item)"

This statement did not add any value to my life.
 
Let argue over this purely from a data structure and math standpoint. Its quite evident you werent implementing a cartesian product. In your test case what is [1,2,3] and what is [a,b]? Are they rows or columns? If so why are they of unequal length. Are the label's to rows and columns ? If so whats the contract exposed to caller's? Please explain in plain lucid english what is the data structure representing in your inputs.

lfarah:
I agree with @Khalian, but I don't agree on the way he said it.

In my opinion, we as an open source community need to give the chance and support to the developers who are still learning because we weren't born knowing everything. Let's write comments that are constructive and not destructive, changing terms like "This statement did not add any value to my life." to "Can you please explain what's going on here?".

@Khalian I totally understand your frustration. You currently work on the repo more than anyone here and may think that maybe there's a lack of responsibility. Also please don't take it personally. I only want the best of the repo and everyone here, so we can work together and grow as developers.

@vilapuigvila To be fair, I do agree with @Khalian and I really don't know what's going on in this PR.
For an example, take a look at #375 and see how we had an issue on which we discussed the feature and how we'd implement it, then @Khalian opened a PR with clean & clear variable names. This extension is denser, so maybe adding some comments would be helpful

PR #392, opened 2017-02-20
<Description> : Removing foreach since it conflicts with the std libr…

Khalian:
…ary.

<Type> : fix

## Checklist

- [ ] New Extension
- [ ] New Test

### DIFF
--- a/Sources/ArrayExtensions.swift
+++ b/Sources/ArrayExtensions.swift
@@ -232,24 +232,11 @@ extension Array where Element: Equatable {
     public mutating func removeFirstObject(_ object: Element) {
         removeFirst(object)
     }
-
-    /// EZSE: Removes all occurrences of the given object
-    @available(*, deprecated: 1.8, renamed: "removeAll(_:)")
-    public mutating func removeObjects(_ objects: Element...) {
-        objects.forEach { self.removeAll($0) }
-    }
-
 }
 
 // MARK: - Deprecated 1.7
 
 extension Array {
-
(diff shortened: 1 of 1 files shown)


Khalian:
Done.

DanielAsher:
Hi, 
I believe that removal of https://github.com/goktugyil/EZSwiftExtensions/blob/928d825dcc5002f468d8cfc0b94b98a5910f160c/Sources/ArrayExtensions.swift#L135 is required for correctly implementing this PR

Khalian:
That method is not deprecated, and we need to think this through if we do want to deprecate it.

DanielAsher:
@Khalian: thanks for this info. I'll go ahead and fork. Probably for the best :) I'd be glad to start making the minor changes for protocol based extensions on concrete types there.

Khalian:
@lfarah @goktugyil Can we deprecate the method that @DanielAsher mentioned : https://github.com/goktugyil/EZSwiftExtensions/blob/928d825dcc5002f468d8cfc0b94b98a5910f160c/Sources/ArrayExtensions.swift#L135 . If so I can put out a minor release adding this to something like 1.9.1.

Khalian:
@DanielAsher As for unblocking you, I will quickly discuss with the other maintainers and maybe put out a minor version. Sorry for the inconvenience.

DanielAsher:
@Khalian thanks so much for your attention to this. maintaining a well-used library has many complexities - best of luck and wishes for the future!

goktugyil:
@Khalian Why are we deprecating it again?

PR #394, opened 2017-03-06
<Description> : Adding a parallel map functionality for collections.

Khalian:
<Type> : feature/test

## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/Sources/CollectionsExtension.swift
+++ b/Sources/CollectionsExtension.swift
@@ -1,0 +1,42 @@
+//
+//  CollectionsExtension.swift
+//  EZSwiftExtensions
+//
+//  Created by Sanyal, Arunav on 3/5/17.
+//  Copyright © 2017 Goktug Yilmaz. All rights reserved.
+//
+
+import Foundation
+
+extension Collection {
+
+#if os(iOS)
+    // EZSE : A parralelized map for collections, operation is non blocking
+    public func parallelizedMap<R>(_ each: @escaping (Self.Iterator.Element) -> R) -> [R?] {
+        let indices = indicesArray()
+        var res = [R?](repeating: nil, count: indices.count)
+        
+        let queue = OperationQueue()
+        let operations = indices.enumerated().map { index, elementIndex in
+            return BlockOperation {
+                res[index] = each(self[elementIndex])
+            }
+        }
+        queue.addOperations(operations, waitUntilFinished: true)
+        
+        return res
+    }
+
+    
+    /// EZSE : Helper method to get an array of collection indices
+    private func indicesArray() -> [Self.Index] {
+        var indicesArray: [Self.Index] = []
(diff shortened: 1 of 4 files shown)


Sources/CollectionsExtension.swift
```diff
@@ -0,0 +1,40 @@
...
+//  Created by Sanyal, Arunav on 3/5/17.
+//  Copyright © 2017 Goktug Yilmaz. All rights reserved.
+//
+
+import Foundation
+
+extension Collection {
+    
+    // EZSE : A parralelized map for collections, operation is non blocking
+    public func pmap<R>(_ each: @escaping (Self.Iterator.Element) -> R) -> [R?] {
```
Khalian: My only callout here would be should the return type be [R?] or [R]. I can go either way. I am not sure as to the safety of these ops though so I kept it as [R?] (e.g. if each method returns a bunch of nils). However [R] is more consumer friendly.

Khalian:
Not sure why this is failing. It was working fine when I tested in locally.

Khalian:
@lfarah whenever you get time.

lfarah:
@Khalian let's merge it and see if it still fails after that

Khalian:
@lfarah I am skeptical about merging and subsequently failing. I will investigate why this is happening.

Khalian:
Man I don't understand why the tests are failing in travis but not on XCode. Can anyone help me out here ?

Khalian:
@piv199 @goktugyil Would you guys mind taking a look at this? This is a pretty non trivial extension.

Sources/CollectionsExtension.swift
```diff
@@ -0,0 +1,42 @@
...
+//  Copyright © 2017 Goktug Yilmaz. All rights reserved.
+//
+
+import Foundation
+
+extension Collection {
+
+#if os(iOS)
+    // EZSE : A parralelized map for collections, operation is non blocking
+    public func pmap<R>(_ each: @escaping (Self.Iterator.Element) -> R) -> [R?] {
```
lfarah: I didn't like the name `pmap` for a method since It's not camel cased and it's not descriptive enough. Maybe `parallelizedMap()`?

Sources/CollectionsExtension.swift
```diff
@@ -0,0 +1,42 @@
...
+//  Created by Sanyal, Arunav on 3/5/17.
+//  Copyright © 2017 Goktug Yilmaz. All rights reserved.
+//
+
+import Foundation
+
+extension Collection {
+
+#if os(iOS)
+    // EZSE : A parralelized map for collections, operation is non blocking
```
lfarah: Typo.
`parralelized` -> `parallelized`

Khalian: Done

PR #396, opened 2017-03-12
UIStackView extensions

kirakik:
- Added UIStackView extensions to add arranged subviews and convenience initializer
- Added tests for UIStackView extension

- [x] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensions/UIStackViewExtensions.swift
+++ b/EZSwiftExtensions/UIStackViewExtensions.swift
@@ -1,0 +1,42 @@
+//
+//  UIStackViewExtensions.swift
+//  EZSwiftExtensionsExample
+//
+//  Created by Kenza Iraki on 2017-03-12.
+//  Copyright © 2017 Goktug Yilmaz. All rights reserved.
+//
+
+import UIKit
+
+@available(iOS 9.0, *)
+extension UIStackView {
+
+
+	/// EZSwiftExtensions: Convenience initializer in order to directly pass the distribution, alignment, axis and spacing of the stack view.
+	///
+	/// - Parameters:
+	///   - distribution: the distribution of the stack view
+	///   - alignment: the alignment of the stack view
+	///   - axis: the axis (e.g. horizontal or vertical)
+	///   - spacing: spacing between subviews, default is 0
+	public convenience init(distribution: UIStackViewDistribution,
+	                        alignment: UIStackViewAlignment,
+	                        axis: UILayoutConstraintAxis,
+	                        spacing: CGFloat = 0) {
+		self.init()
+		self.distribution = distribution
+		self.alignment = alignment
+		self.axis = axis
+		self.spacing = spacing
+	}
+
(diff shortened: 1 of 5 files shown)


Khalian:
Would you mind squashing you changelog commit and actual code commit into one single commit ?

EZSwiftExtensions/UIStackViewExtensions.swift
```diff
@@ -0,0 +1,42 @@
...
+		self.distribution = distribution
+		self.alignment = alignment
+		self.axis = axis
+		self.spacing = spacing
+	}
+
+	/// EZSwiftExtensions: Add multiple subviews at once. They will be added in order (top to bottom for a vertical stack view and left to right for a vertical stack view)
+	///
+	/// - Parameter views: array of views to add
+	public func addArrangedSubviews(_ views: [UIView]) {
```
Khalian: Suggestion : varargs instead of array?

kirakik: Good point, I was basing this off of the UIViewExtension method `addSubviews` which takes an array and not varargs. I was thinking of updating that but I'm guessing it might be breaking changes for those already using it?

kirakik: It also could cause some memory leaks seeing as it contains a `forEach` capturing `self` strongly

``` swift
    public func addSubviews(_ views: [UIView]) {
        views.forEach { eachView in
            self.addSubview(eachView)
        }
    }
```

Khalian: As long as they are in different namespaces it should be fine. If not lets keep it this way then.

kirakik:
@Khalian I'm having trouble squashing my commits, every time I try it just seems to add more commits rather than squash them... There is an option on github to squash and merge id that's what you're worried about

EZSwiftExtensions/UIStackViewExtensions.swift
kirakik: I understand, but think there could be some improvements there, especially when it comes to the potential/likely memory leak this could cause

Khalian:
@kirakik git rebase -i HEAD~n where n is some big number (say 10). This will open up an interactive rebase. Then replace the word "pick" with "f" for the last k - 1 commits (where k is the total number of commits that you have for this PR). This will rewind and replay all k commits into 1. Then do a force push on your branch. The PR will subsequently pick up all content changes.

kirakik:
Thank you for the explanation @Khalian, I was doing it correctly but wasn't force pushing my changes. It's now done

Khalian:
I approved your changes. We have a policy in which 2 collaborators approvals are needed prior to a merge. @lfarah would you mind taking a look at this ?

lfarah:
Hey @kirakik, nice job! Thanks for contributing 👍

lfarah:
Hey @Khalian, we changed the CHANGELOG for this commit, but we already released `1.9`.

Khalian:
@lfarah Ah crap should have seen that. i will fix it now.

PR #397, opened 2017-03-13
UIView extension memory leak fix

kirakik:
Fixed potential memory leak of addSubviews UIView extension method

- [x] New Extension

### DIFF
--- a/Sources/UIViewExtensions.swift
+++ b/Sources/UIViewExtensions.swift
@@ -31,8 +31,8 @@ extension UIView {
 
     /// EZSwiftExtensions, add multiple subviews
     public func addSubviews(_ views: [UIView]) {
-        views.forEach { eachView in
-            self.addSubview(eachView)
+        views.forEach { [weak self] eachView in
+            self?.addSubview(eachView)
         }
     }
 


Sources/UIViewExtensions.swift
```diff
@@ -31,8 +31,8 @@ extension UIView {
 
     /// EZSwiftExtensions, add multiple subviews
     public func addSubviews(_ views: [UIView]) {
-        views.forEach { eachView in
-            self.addSubview(eachView)
+        views.forEach { [weak self] eachView in
```
Khalian: Good catch. I presume this implies the subview has a reference to the parent view and vice versa and due to ARC (Automated ref counting) there is no way to resolve the ref cycle without a weak ref.

kirakik: Exactly. I was going to change it for a `for...in...` loop, but I decided against it, didn't want to change too much here

Khalian:
@kirakik You can change the internals without breaking the API contract. As long as all unit tests for this pass through, it should be fine (although I dont remember if it has any).

Sources/UIViewExtensions.swift
furuyan: `forEach` is not `@escaping` method.  So i think this case is not need to describe `weak self`.

`public func forEach(_ body: (Element) throws -> Void) rethrows`

goktugyil: @furuyan What exactly are you recommending?

PR #398, opened 2017-03-14
Fixed enum cases

furuyan:
Enum cases should be lower camel case in Swift3.

## Checklist

- [ ] New Extension
- [ ] New Test

### DIFF
--- a/EZSwiftExtensionsTests/UIFontTests.swift
+++ b/EZSwiftExtensionsTests/UIFontTests.swift
@@ -13,7 +13,7 @@ class UIFontTests: XCTestCase {
     
     let TEST_SIZE = CGFloat(10)
     
-    let TEST_FONT_TYPE = FontType.Bold
+    let TEST_FONT_TYPE = FontType.bold
     
     override func setUp() {
         super.setUp()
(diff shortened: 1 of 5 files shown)


EZSwiftExtensionsTests/UIFontTests.swift
```diff
@@ -13,7 +13,7 @@ class UIFontTests: XCTestCase {
     
     let TEST_SIZE = CGFloat(10)
     
-    let TEST_FONT_TYPE = FontType.Bold
+    let TEST_FONT_TYPE = FontType.bold
```
Khalian: Can I get a reference to the code standards which suggest this change ?

furuyan: Sure.

https://swift.org/documentation/api-design-guidelines/#follow-case-conventions
This guideline says `Names of types and protocols are UpperCamelCase. Everything else is lowerCamelCase.`

swift-evolution: Decision Notes
https://lists.swift.org/pipermail/swift-evolution-announce/2016-March/000053.html

lfarah: what about ```TEST_FONT_TYPE```? Shouldn't it be ```testFontType```?

furuyan: @lfarah Yes. Exactly!
But should i include this change in this PR?

Khalian: Yea I guess you can. Its essentially refactoring, feel free to update the PR. I am approving this for now.

Khalian:
I am a little worried about breaking userspace. Are the enums public accessible ?

furuyan:
You are right. These enums are public accessible.
I should write CHANGELOG, right?

Khalian:
Yes and also facilitate a deprecation path. 

Lets keep both sets and mark the old ones as deprecated.

furuyan:
I got it.

furuyan:
I tried to implement. 
But compiler says, enum cases must be unique.

Do you guys have any ideas?

Khalian:
I see, the values itself must be unique. I.E. case Regular = "Regular" and case regular = "regular" is legal since the values differ on the first letter being uppercase vs lowercase. 

I am not sure what should be the best course of action here. @lfarah My only concern is breaking userspace.

Khalian:
@furuyan Do you wish to continue with this PR?

furuyan:
I can't think of any good solution. So close this PR for now.

PR #403, opened 2017-04-01
Fix DBL_MAX deprecation warning

ayn:
## Checklist
- [ ] New Extension
- [ ] New Test

### DIFF
--- a/Sources/StringExtensions.swift
+++ b/Sources/StringExtensions.swift
@@ -402,7 +402,7 @@ extension String {
             paragraphStyle.lineBreakMode = lineBreakMode!
             attrib.updateValue(paragraphStyle, forKey: NSParagraphStyleAttributeName)
         }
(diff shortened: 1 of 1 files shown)


Sources/StringExtensions.swift
```diff
@@ -402,7 +402,7 @@ extension String {
             paragraphStyle.lineBreakMode = lineBreakMode!
             attrib.updateValue(paragraphStyle, forKey: NSParagraphStyleAttributeName)
         }
-        let size = CGSize(width: width, height: CGFloat(DBL_MAX))
+        let size = CGSize(width: width, height: CGFloat(Float.greatestFiniteMagnitude))
```
Khalian: Are these values equivalent in numerical terms ?

ayn: Just checked, `Double.greatestFiniteMagnitude` is the same, just committed a change. Thanks.

PR #411, opened 2017-04-22
Fixed broken Array extension get(at:) (+test)

dfrib:
The `get(at:)` extension previously led to a runtime exception when being called from empty arrays (`[]`), as the constructed local `ClosedRange` `lowerBound...upperBound` yielded the invalid `ClosedRange` `0...-1` for most supplied ranges.

Also extended the `testGet_SubArray()` test of `ArrayTests` to test the extension on empty arrays (which would've caught the case above).

Swiftlint @ ArrayExtensions OK.

I'm somewhat uncertain as to what boxes that apply in the checklist below for this commit, as I haven't added a new test, but have modified one. I'm marking it as trivial for now (refactoring/slightly extended test), but please advice me (/edit this msg) if this should be changed.

## Checklist

- [ ] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -21,11 +21,16 @@ class ArrayTests: XCTestCase {
     }
     
     func testGet_SubArray() {
-        setUp()
         XCTAssertEqual(numberArray.get(at: 2...4), [2,3,4])
         XCTAssertEqual(numberArray.get(at: -1...8), [0,1,2,3,4,5,1])
         XCTAssertEqual(numberArray.get(at: -1...3), [0,1,2,3])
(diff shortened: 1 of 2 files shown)


dfrib:
Happy to help.

I also caught a similar crash case for the `random()` method of `Dictionary` (for empty dict), but I don't know how to fix that straight off the bat without changing the return type to `Optional`, which I guess would be a breaking change. Posted that one as [Issue 413](https://github.com/goktugyil/EZSwiftExtensions/issues/413).

Khalian:
Yea I looked at that. Changing the return type to optional might break the contract of what we expose to the users (in hindsight, I never expected a random to not return something from the list - which was my fault). 

I will try to implement your first option. 

Right now I am more focussed on getting swiftlint back online. Travis misbehaves like this all the time.

PR #414, opened 2017-04-22
Minor refactoring of Dictionary extensions

dfrib:
Minor refactoring of three dict extensions. I could omit the `testAll(_:)` refactoring, which now makes use of functional methods rather than explicit classic loops, but at the cost of using negative logic (in two senses), possibly worsening readability. Thoughts?

A related comment:
* I might be wrong, but it seems as if the `toArray(_:)` method duplicates the functionality of [the `map(:_)` method of Dictionary](https://developer.apple.com/reference/swift/dictionary/1689101-map), available via [the default implementation from 'Collection'](https://github.com/apple/swift/blob/83908723d5f7896a9f2a5133ea67b20f5efe6240/stdlib/public/core/Collection.swift#L1404). This fact is quite apparent after the re-factoring of this commit, and possibly prompts discussing a future deprecation of the `toArray(_:)` method, unless it's deemed valuable as a utility method for pure semantic reasons (however breaking change, so nothing for now).

Swiftlint @ DictionaryExtensions OK (unchanged)

## Checklist

- [ ] New Extension
- [ ] New Test


Sources/DictionaryExtensions.swift
```diff
@@ -19,7 +19,7 @@ extension Dictionary {
         var result = self
         dictionaries.forEach { (dictionary) -> Void in
             dictionary.forEach { (key, value) -> Void in
```
Khalian: Can this be further converted via a map method ?

dfrib: I did look this over, but, at the top of my head, I couldn't find a better approach. Maybe once the accepted [SE-0165 (Dictionary & Set Enhancements)](https://github.com/apple/swift-evolution/blob/master/proposals/0165-dict.md) is implemented, one could make use of merging initializer to refactor it.

PR #416, opened 2017-04-22
Added two overloads for the 'removeAll(_:)' extension of 'Array'

dfrib:
**Summary**
Added two overloads for the `removeAll(_:)` extension of `Array`:
* One for an array parameter rather than a variadic one.
* Another one for an array parameter but for a more specific overload; elements conforming to `Hashable`.

Updated the existing variadic overload to simply call the new array one.

Extended the existing test `testRemoveObjects()` into two: one testing the
single variadic overload and one testing the two new array overloads.

Swiftlint @ ArrayExtensions OK.

Please advise me if I'm missing some PR best practices/guidelines for EZSE.

**Background & Motivation**
The existing implementation will for similarily growing size of `self` and the supplied variadic parameter have `O(n^2)` complexity. This isn't any limitation at all for the variadic overload, as the number of supplied elements will be limited ("constant"). For the new array overload, however, a custom-sized array can be supplied as a parameter, in which case it could be useful to make use of the (amortized) `O(1)` lookup for `Hashable` elements in a `HashedCollection` (e.g. `Set`). The `Hashable` overload will, for each element in `self`, perform an (amortized) `O(1)` lookup for the elements in the supplied array, at the cost of a single `Set` instantation of the array parameter, such that the `Hashable` overload will have `O(n)` complexity (for similarily growing size of `self` and supplied array parameter).

**Impact**
There should be no usage impact on existing use of the `removeAll(_:)` method: only a new array parameter overload available for use.

## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/Sources/ArrayExtensions.swift
+++ b/Sources/ArrayExtensions.swift
@@ -132,13 +132,15 @@ extension Array where Element: Equatable {
         self.remove(at: index)
     }
 
-    // EZSE: Removes all occurrences of the given object
+    /// EZSE: Removes all occurrences of the given object(s)
     public mutating func removeAll(_ elements: Element...) {
-        for element in elements {
-            for index in self.indexes(of: element).reversed() {
-                self.remove(at: index)
-            }
-        }
+        removeAll(elements)
+    }
+
+    /// EZSE: Removes all occurrences of the given object(s)
+    public mutating func removeAll(_ elements: [Element]) {
+        // COW ensures no extra copy in case of no removed elements
+        self = filter { !elements.contains($0) }
     }
 
     /// EZSE: Difference of self and the input arrays.
@@ -198,6 +200,16 @@ extension Array where Element: Equatable {
     /// EZSE: Returns an array consisting of the unique elements in the array
     public func unique() -> Array {
         return reduce([]) { $0.contains($1) ? $0 : $0 + [$1] }
+    }
+}
+
+extension Array where Element: Hashable {
+
+    /// EZSE: Removes all occurrences of the given object(s)
+    public mutating func removeAll(_ elements: [Element]) {
+        let elementsSet = Set(elements)
+        // COW ensures no extra copy in case of no removed elements
+        self = filter { !elementsSet.contains($0) }
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ All notable changes to this project will be documented in this file.
    - `addArrangedSubviews(_ views: UIView...)` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/396) by *kirakik*
    - `init(distribution: UIStackViewDistribution, alignment: UIStackViewAlignment, axis: UILayoutConstraintAxis, spacing: CGFloat)` in [[PR]](https://github.com/goktugyil/         EZSwiftExtensions/pull/396) by *kirakik*
 
+### Added extensions
+1. **Array**
+   - `removeAll(_ elements: [Element])` (new overload) for `Equatable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
+   - `removeAll(_ elements: [Element])` (new overload) for `Hashable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
+
 ### Deprecated/Renamed extensions
 1. **UIViewController**
    - `public func hideKeyboardWhenTappedAround(cancelTouches: Bool = false)` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/395) by *lfarah*
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -63,13 +63,46 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(numberArray, compareArray)
     }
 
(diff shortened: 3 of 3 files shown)


Sources/ArrayExtensions.swift
```diff
@@ -132,13 +132,15 @@ extension Array where Element: Equatable {
         self.remove(at: index)
     }
 
-    // EZSE: Removes all occurrences of the given object
+    /// EZSE: Removes all occurrences of the given object(s)
     public mutating func removeAll(_ elements: Element...) {
-        for element in elements {
```
dfrib: (This body refactoring is already present in open PR #412, included here to ease possible merge given that both PR are accepted / reviewed and revised)

Sources/ArrayExtensions.swift
```diff
@@ -132,13 +132,15 @@ extension Array where Element: Equatable {
...
-            for index in self.indexes(of: element).reversed() {
-                self.remove(at: index)
-            }
-        }
+        removeAll(elements)
+    }
+
+    /// EZSE: Removes all occurrences of the given object(s)
+    public mutating func removeAll(_ elements: [Element]) {
```
Khalian: So is this a counter to https://bugs.swift.org/browse/SR-128 ?

Sources/ArrayExtensions.swift
```diff
@@ -201,6 +203,16 @@ extension Array where Element: Equatable {
     }
 }
 
+extension Array where Element: Hashable {
+
+    /// EZSE: Removes all occurrences of the given object(s)
+    public mutating func removeAll(_ elements: [Element]) {
+        let elementsSet = Set(elements)
```
Khalian: This operation is a bag transformation to a set. E.g [1, 1, 2, 3, 2] would become [1, 2, 3]

I am not inclined towards the contract of removeAll to also include a bag -> set transform.

Khalian:
If you are proposing these overloads, would you mind adding CHANGELOG entries for the same ?

Sources/ArrayExtensions.swift
dfrib: Yes, you could say that (due to that limitation). We supply an additional overload to allow users to call `removeAll(_:)` with variadic as well as array arguments.

Khalian: Sigh, splatting should be a given in modern programming languages. Oh well.

Sources/ArrayExtensions.swift
dfrib: The bag transformation is not visible to the user: we only make use of the bag transformation locally for the supplied array argument (the one that tells us which elements in `self` that are to be removed), to allow `O(1)` lookup in the `filter` predicate below (`contains` applied to a `Set` instance rather than an `Array` instance). Since the supplied `elements` array are just identifiers we needn't duplicate them for the `self` removal lookup, even if the callee sends us an `Array` that contains duplicate elements. The result of the call to `removeAll(_:)` should be the same, but `O(n)` for `Hashable` objects and `O(n^2)` for those that are not. But I might've misunderstood your comment, please correct me if I'm wrong!

dfrib:
Ah, of course! Do I add the new overloads under _Unreleased_ in the changelog?

Sources/ArrayExtensions.swift
Khalian: So lets assume the array : val arr = [1, 1, 2, 2, 3] where the elements are hashable. Now, say elements Set = {1, 2, 3}. arr.removeAll(3) would return [1, 2]. 

But the expected contract was [1, 1, 2, 2].

dfrib: I might still be overlooking something, but note that the `Array` to `Set` transformation is never applied to the caller (i.e., `self`), it is only applied to the supplied `elements` array, as there is no value having duplicates in the array that describes which elements of `self` that are to be removed. For the caller (`self`), an `O(n)` filter is applied onto the `self` `Array`, and only the predicate in the `filter` closures makes use of `Hashable` logic for the supplied `elements`.

To show in practice, the following addition to `testRemoveObjectsByArray()` would cover your example above, and asserts to `true`:

```
var numberArrayWithDupes = [1, 1, 2, 2, 3]
let compareArrayForDupes = [1, 1, 2, 2]
numberArrayWithDupes.removeAll([3]) // use Hashable's removeAll(_:[Elements])
XCTAssertEqual(numberArrayWithDupes, compareArrayForDupes)
```

Finally, note that calling `removeAll(_:)` by variadic arguments (e.g. `arr.removeAll(3)` as in your comment above) will always call the single `Equatable` variadic overload. As I write in the description above, a call with variadic arguments will always be limited in size, in which case making using of `O(1)` lookup for `Hashable` elements, specifically, have limited value.

Khalian: Oh my bad, so the filter is on the self itself ? (I am used to filters as method calls).

dfrib: Exactly, the `filter` applied upon `self` is `O(n)` (for the `Hashable` as well as for the `Equatable` overload), but given `Hashable` elements, we can avoid the nested predicate in the `filter` closure (namely the `contains` call) from also being "`O(n)`" (here w.r.t. scaling size of the input argument `elements`; I guess we should call it `O(m)` :) ).

Could be a corner-case optimization, but for similarly scaling size of `self` and `elements` (when supplied as an array), `Hashable` elements would have `O(n)` scaling for the call (`O(n)` filter and `O(1)` `contains` predicate) whereas non-`Hashable` elements would be `O(n^2)` (`O(n)` filter and `O(n)` `contains` predicate).

PR #419, opened 2017-05-07
Adding support for checking if app is running in testflight mode.

Khalian:
## Checklist

- [x] New Extension
- [ ] New Test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_script:
 script:
   - set -o pipefail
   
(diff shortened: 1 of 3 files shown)


Khalian:
@goktugyil  Would you mind taking a look at this ?

goktugyil:
You want me to test this inside my app while its in test flight?

Khalian:
Yes, I am not actively writing an App at this time and some friends have asked for this. So that would be helpful.

Also, approving it so that I can merge it.

goktugyil:
I'll check this when I send a TestFlight version, probably 1-2 weeks later.

goktugyil:
Tested this on TestFlight, and returns true as expected.
Tested in normal install, and return false.
Haven't tested on live app, is that test also needed?

Khalian:
@goktugyil No, that should be good enough.

PR #424, opened 2017-05-20
Adding round corners to UIView

Vic-L:
## Checklist

- [ ] New Extension
- [x] New Test

**Gist**
- Overload `roundView()` with one that takes in parameters to add colored border to the view
- New method `nakedView()` to remove all masks of UIView

*PS its my first time contributing to a open source project. Thank you for your guidance 🙇 *

### DIFF
--- a/EZSwiftExtensionsTests/UIViewTests.swift
+++ b/EZSwiftExtensionsTests/UIViewTests.swift
@@ -60,5 +60,56 @@ class UIViewTests: XCTestCase {
         view.removeSubviews()
         XCTAssertEqual(view.subviews.count, 0)
     }
+    
+    func testRoundView() {
+        let wideView = UIView(x: 0, y: 0, w: 200, h: 80)
+        XCTAssertEqual(wideView.layer.cornerRadius, 0)
+        XCTAssertEqual(wideView.layer.borderWidth, 0)
+        XCTAssertNotEqual(wideView.layer.borderColor, UIColor.clear.cgColor)
+        
+        wideView.roundView()
+        XCTAssertEqual(wideView.layer.cornerRadius, CGFloat(40.0))
+        XCTAssertEqual(wideView.layer.borderWidth, 0)
+        XCTAssertEqual(wideView.layer.borderColor, UIColor.clear.cgColor)
+        
+        let tallView = UIView(x: 0, y: 0, w: 30, h: 300)
+        XCTAssertEqual(tallView.layer.cornerRadius, 0)
+        XCTAssertEqual(tallView.layer.borderWidth, 0)
+        XCTAssertNotEqual(tallView.layer.borderColor, UIColor.clear.cgColor)
+        
+        tallView.roundView()
+        XCTAssertEqual(tallView.layer.cornerRadius, CGFloat(15.0))
+        XCTAssertEqual(tallView.layer.borderWidth, 0)
+        XCTAssertEqual(tallView.layer.borderColor, UIColor.clear.cgColor)
+    }
+    
+    func testRoundViewWithBorder() {
+        let view = UIView(x: 0, y: 0, w: 100, h: 100)
+        let color = UIColor.blue
+        let borderWidth: CGFloat = 5.0
+
+        XCTAssertEqual(view.layer.cornerRadius, 0)
+        XCTAssertEqual(view.layer.borderWidth, 0)
+        XCTAssertNotEqual(view.layer.borderColor, color.cgColor)
+        
+        view.roundView(withBorderColor: color, withBorderWidth: borderWidth)
+        XCTAssertEqual(view.layer.cornerRadius, CGFloat(50.0))
+        XCTAssertEqual(view.layer.borderWidth, borderWidth)
+        XCTAssertEqual(view.layer.borderColor, color.cgColor)
+    }
+    
+    func testNakedView() {
+        let view = UIView(x: 0, y: 0, w: 100, h: 100)
+        let color = UIColor.blue
+        let borderWidth: CGFloat = 5.0
+        
+        view.roundView(withBorderColor: color, withBorderWidth: borderWidth)
+        XCTAssertEqual(view.layer.cornerRadius, CGFloat(50.0))
+        XCTAssertEqual(view.layer.borderWidth, borderWidth)
+        
+        view.nakedView()
+        XCTAssertEqual(view.layer.cornerRadius, CGFloat(50.0))
+        XCTAssertEqual(view.layer.borderWidth, 0)
+    }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/UIViewExtensions.swift
```diff
@@ -565,9 +565,34 @@ extension UIView {
...
     public func roundView() {
-        self.layer.cornerRadius = min(self.frame.size.height, self.frame.size.width) / 2
+        self.setCornerRadius(radius: min(self.frame.size.height, self.frame.size.width) / 2)
+        self.layer.borderWidth = 0
+    }
+    
+    /**
+     EZSwiftExtensions - Mask square/rectangle UIView with a circular/capsule cover, with a border of desired color and width around it
+     
+     - Parameter withBorderColor: Color of circuler border
```
Khalian: We dont need these comments. The args are seld documenting.

Sources/UIViewExtensions.swift
```diff
@@ -565,9 +565,34 @@ extension UIView {
         self.layer.mask = mask
     }
 
-    /// EZSwiftExtensions
+    /**
+     EZSwiftExtensions - Mask square/rectangle UIView with a circular/capsule cover
```
Khalian: We can put this comment on a single line.

Sources/UIViewExtensions.swift
```diff
@@ -565,9 +565,34 @@ extension UIView {
...
+        self.layer.borderWidth = 0
+    }
+    
+    /**
+     EZSwiftExtensions - Mask square/rectangle UIView with a circular/capsule cover, with a border of desired color and width around it
+     
+     - Parameter withBorderColor: Color of circuler border
+     - Parameter withBorderWidth: Width of circuler border
+     */
+    public func roundView(withBorderColor color: UIColor, withBorderWidth width: CGFloat) {
```
Khalian: I am thinking if color can be an optional parameter. That way roundView above can become a special case on the roundView with optional color and borderwidth = 0. Besides, a programmer can choose to have no color but some border width.

Sources/UIViewExtensions.swift
```diff
@@ -565,9 +565,34 @@ extension UIView {
...
+        self.setCornerRadius(radius: min(self.frame.size.height, self.frame.size.width) / 2)
+        self.layer.masksToBounds = true
```
Khalian: I am not aware what masks to bound does.

EZSwiftExtensionsTests/UIViewTests.swift
```diff
@@ -60,5 +60,58 @@ class UIViewTests: XCTestCase {
...
+        tallView.roundView()
+        XCTAssertEqual(tallView.layer.cornerRadius, CGFloat(15.0))
+        XCTAssertEqual(tallView.layer.masksToBounds, true)
+    }
+    
+    func testRoundViewWithBorder() {
+        let view = UIView(x: 0, y: 0, w: 100, h: 100)
+        let color = UIColor.blue
+        let borderWidth: CGFloat = 5.0
+        view.layer.borderColor = nil // override default view.layer.borderColor
```
Khalian: I am not wholly comfortable with this nil set and then assert. Maybe consider optional bordercolor as arg or something. Then asset the content of the optional to find nothing there.

Sources/UIViewExtensions.swift
```diff
@@ -565,9 +565,34 @@ extension UIView {
...
+        self.layer.borderWidth = width
+        self.layer.borderColor = color.cgColor
+    }
+    
+    /**
+     EZSwiftExtensions - Remove all masking around UIView
```
Khalian: This comment can be in one line.

Sources/UIViewExtensions.swift
```diff
@@ -565,9 +565,34 @@ extension UIView {
...
+     */
+    public func nakedView() {
```
Khalian: Whats the use case for this?

Khalian:
Could you also add a changelog entry and squash your commits together?

Sources/UIViewExtensions.swift
Vic-L: Ah I used it for a rounded UIImageView showing a user's profile picture.

When user logs in and his/her profile picture is loaded, the mask needs to be on to make the image rounded, and with a border around it.
When user logs out, the image will change to a image asset, and the mask needs to be taken out.

I thought a helper method to remove any mask present would be helpful.

Sources/UIViewExtensions.swift
Vic-L: I was under the impression that the mask will not resize along with the view is masksToBounds is false. Apparently I'm wrong :( I'll remove it

This is the code snippet that i used to experiment
```
override func viewDidLoad() {
        let view = UIView(frame: CGRect(x: 0, y: 0, w: 250, h: 250))
        view.roundView()
        view.backgroundColor = .green
        view.layer.masksToBounds = false // toggle here
        
        self.view.addSubview(view)
        if #available(iOS 10.0, *) {
            timer = Timer.scheduledTimer(withTimeInterval: 0.5, repeats: true, block: { (_timer) in
                view.snp.updateConstraints({ (make) in
                    make.height.equalTo(ez.screenHeight)
                    make.width.equalTo(ez.screenWidth)
                })
                
                self.view.setNeedsLayout()
                
                UIView.animate(withDuration: 2, animations: {
                    self.view.layoutIfNeeded()
                }, completion: { (completed) in
                    if completed {
                        view.snp.updateConstraints({ (make) in
                            make.height.equalTo(100)
                            make.width.equalTo(100)
                        })
                        
                        self.view.setNeedsLayout()
                    }
                    
                    UIView.animate(withDuration: 2, animations: {
                        self.view.layoutIfNeeded()
                    })
                })
            })
        } else {
            // Fallback on earlier versions
        }
        
    }
```

Vic-L:
oops i think i did the wrong thing
`git push -f` ~= `squash` eh :(

EZSwiftExtensionsTests/UIViewTests.swift
```diff
@@ -60,5 +60,56 @@ class UIViewTests: XCTestCase {
...
+        XCTAssertEqual(tallView.layer.borderWidth, 0)
+        XCTAssertNotEqual(tallView.layer.borderColor, UIColor.clear.cgColor)
+        
+        tallView.roundView()
+        XCTAssertEqual(tallView.layer.cornerRadius, CGFloat(15.0))
+        XCTAssertEqual(tallView.layer.borderWidth, 0)
+        XCTAssertEqual(tallView.layer.borderColor, UIColor.clear.cgColor)
+    }
+    
+    func testRoundViewWithBorder() {
```
Khalian: Minor : I think you can call this method testRoundViewWithBorderAndColor.

Vic-L:
👌 
Many thx for your patience!

Great work on the pod btw!

PR #425, opened 2017-05-20
feat: convert double to CGFloat conveniently

Vic-L:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ All notable changes to this project will be documented in this file.
    - `removeAll(_ elements: [Element])` (new overload) for `Hashable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
 
 2. **Double**
-   - `var toCGFloat` in PR by *Vic-L*
+   - `var toCGFloat` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/425) by *Vic-L*
 
 ### Deprecated/Renamed extensions
 1. **UIViewController**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file.
    - `removeAll(_ elements: [Element])` (new overload) for `Equatable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
    - `removeAll(_ elements: [Element])` (new overload) for `Hashable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
 
+2. **Double**
+   - `var toCGFloat` in PR by *Vic-L*
+
 ### Deprecated/Renamed extensions
 1. **UIViewController**
    - `public func hideKeyboardWhenTappedAround(cancelTouches: Bool = false)` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/395) by *lfarah*
--- a/EZSwiftExtensionsTests/DoubleTests.swift
+++ b/EZSwiftExtensionsTests/DoubleTests.swift
@@ -45,10 +45,7 @@ extension DoubleTests {
     func testCGFloat() {
         XCTAssertEqual(double.toCGFloat, 3.14)
         XCTAssertEqual(otherDouble.toCGFloat, -147.9564)
-        XCTAssertTrue(double.toCGFloat is CGFloat)
-        XCTAssertTrue(otherDouble.toCGFloat is CGFloat)
-        XCTAssertFalse(double.toCGFloat is Double)
-        XCTAssertFalse(otherDouble.toCGFloat is Double)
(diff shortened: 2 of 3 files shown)


Sources/DoubleExtensions.swift
```diff
@@ -14,6 +14,9 @@ extension Double {
 
     /// EZSE: Converts Double to Int
     public var toInt: Int { return Int(self) }
+    
+    /// EZSE: Converts Double to CGFloat
+    public var toCGFloat: CGFloat { return CGFloat(self) }
```
Khalian: Is there a possibility of precision loss in this conversion? (E.g. Very large or high number of fraction digits causing some of those to be lost). If yes, we need to write some tests around that to document the edge case.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file.
    - `removeAll(_ elements: [Element])` (new overload) for `Equatable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
    - `removeAll(_ elements: [Element])` (new overload) for `Hashable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
 
+2. **Double**
+   - `var toCGFloat` in PR by *Vic-L*
```
Khalian: I think you need to put the PR number here. You can look at some of the changelog entries to understand how its done.

EZSwiftExtensionsTests/DoubleTests.swift
```diff
@@ -42,6 +42,15 @@ extension DoubleTests {
         XCTAssertNotEqual(Double(anotherDouble.toInt), anotherDouble)
     }
     
+    func testCGFloat() {
+        XCTAssertEqual(double.toCGFloat, 3.14)
+        XCTAssertEqual(otherDouble.toCGFloat, -147.9564)
+        XCTAssertTrue(double.toCGFloat is CGFloat)
```
Khalian: Any particular reason for testing the types ?

CHANGELOG.md
Vic-L: Hi, will like to check
The PR link is only created after a PR is made, so that means the first PR will always fail the TravisCI check...is this expected?

Khalian: I guess you are right. But it has an ID now so I guess you can put it.

EZSwiftExtensionsTests/DoubleTests.swift
Vic-L: The main thing that is changed in this function is the type of the variable, so I thought there's a need to explicitly test it.
The Swift compiler is saying that this is not necessary. Should I remove it?

Khalian: Yes I believe its safe to remove it. Swift is statically and strongly typed for this very reason.

Sources/DoubleExtensions.swift
Vic-L: There will be loss of precision at different values on a 32bit device, which are limited to iPhone5 and before, vs on a 64-bit device

As documented in `CGBase.h`, `CGFloat` is a `typedef` for `double` or `float` depending on the device.
```
#if defined(__LP64__) && __LP64__
# define CGFLOAT_TYPE double
# define CGFLOAT_IS_DOUBLE 1
# define CGFLOAT_MIN DBL_MIN
# define CGFLOAT_MAX DBL_MAX
#else
# define CGFLOAT_TYPE float
# define CGFLOAT_IS_DOUBLE 0
# define CGFLOAT_MIN FLT_MIN
# define CGFLOAT_MAX FLT_MAX
#endif
```

How should we document it here?

Khalian: I guess you can define a similar if def strategy in your unit tests, and then run the worst case values. (DBL MIN, MAX and FLT MIN, MAX). 

But then again I dont remember which device type is default in the build target. Our build targets are only based on platform type (ios vs tvos).

EZSwiftExtensionsTests/DoubleTests.swift
```diff
@@ -42,6 +42,12 @@ extension DoubleTests {
         XCTAssertNotEqual(Double(anotherDouble.toInt), anotherDouble)
     }
     
+    func testCGFloat() {
+        XCTAssertEqual(double.toCGFloat, 3.14)
+        XCTAssertEqual(otherDouble.toCGFloat, -147.9564)
+        XCTAssertEqual(anotherDouble.toCGFloat, 231349.678450342834857392948575)
```
Khalian: I presume this value is DBL MAX ?

Vic-L: This value is obtained from the present test cases. Ill change it to something that is more robust.

Sources/DoubleExtensions.swift
Vic-L: Ill use `greatestFiniteMagnitude` method which will guard against this.

Will this be ok for a test?

PR #426, opened 2017-05-21
Match httpDateString with ISO8601 formats

Vic-L:
## Checklist

- [ ] New Extension
- [ ] New Test

Default timestamp from http json responses for Ruby on Rails follows ISO8601 format.

### DIFF
--- a/EZSwiftExtensionsTests/DateTests.swift
+++ b/EZSwiftExtensionsTests/DateTests.swift
@@ -46,14 +46,18 @@ class DateTests: XCTestCase {
         let fromStartOfMin = TimeInterval(11) // seconds from start of minute
 
         // When
-        let rfc1123 = Date(httpDateString: "Sat, 14 Nov 2015 16:05:11 GMT")
-        let rfc850 = Date(httpDateString: "Saturday, 14-Nov-15 16:05:11 GMT")
-        let asctime = Date(httpDateString: "Sun Nov 14 16:05:11 2015")
-        let iso8601DateOnly = Date(httpDateString: "2015-11-14")
-        let iso8601DateHrMinOnly = Date(httpDateString: "2015-11-14T16:05+00:00")
-        let iso8601DateHrMinSecOnly = Date(httpDateString: "2015-11-14T16:05:11+00:00")
-        let iso8601DateHrMinSecMs = Date(httpDateString: "2015-11-14T16:05:11.123+00:00") // random nanosecond value 123
-        let invalid = Date(httpDateString: "2015-11-14 16:05:11")
+        let correctDateString = "2015-11-14"
+        let correctTimeString = "16:05:11"
+        let correctZoneString = "+00:00"
+        
+        let rfc1123 = Date(httpDateString: "Sat, 14 Nov 2015 \(correctTimeString) GMT")
+        let rfc850 = Date(httpDateString: "Saturday, 14-Nov-15 \(correctTimeString) GMT")
+        let asctime = Date(httpDateString: "Sun Nov 14 \(correctTimeString) 2015")
+        let iso8601DateOnly = Date(httpDateString: correctDateString)
+        let iso8601DateHrMinOnly = Date(httpDateString: "\(correctDateString)T16:05\(correctZoneString)")
+        let iso8601DateHrMinSecOnly = Date(httpDateString: "\(correctDateString)T\(correctTimeString)\(correctZoneString)")
+        let iso8601DateHrMinSecMs = Date(httpDateString: "\(correctDateString)T\(correctTimeString).123\(correctZoneString)") // random nanosecond value 123
+        let invalid = Date(httpDateString: "\(correctDateString) \(correctTimeString)")
 
         // Test
         XCTAssertEqual(rfc1123, self.date)
--- a/EZSwiftExtensionsTests/DateTests.swift
+++ b/EZSwiftExtensionsTests/DateTests.swift
@@ -40,17 +40,26 @@ class DateTests: XCTestCase {
         // Given
         let expectedResult = Date(timeIntervalSince1970: 784_111_777)
         let fromGMT = TimeInterval(NSTimeZone.local.secondsFromGMT())
+        let fromStartOfDay = TimeInterval(8 * 3600 + 49 * 60 + 37) // seconds from start of day
 
         // When
         let rfc1123 = Date(httpDateString: "Sun, 06 Nov 1994 08:49:37 GMT")
         let rfc850 = Date(httpDateString: "Sunday, 06-Nov-94 08:49:37 GMT")
         let asctime = Date(httpDateString: "Sun Nov  6 08:49:37 1994")
+        let iso8601DateOnly = Date(httpDateString: "1994-11-06")
+        let iso8601DateHrMinOnly = Date(httpDateString: "1994-11-06T08:49:37+00:00")
+        let iso8601DateHrMinSOnly = Date(httpDateString: "1994-11-06T08:49:37+00:00")
+        let iso8601DateHrMinSMs = Date(httpDateString: "1994-11-06T08:49:37.000+00:00")
         let invalid = Date(httpDateString: "1994-11-06 08:49:37")
 
         // Test
         XCTAssertEqual(rfc1123, expectedResult)
         XCTAssertEqual(rfc850, expectedResult)
         XCTAssertEqual(asctime?.addingTimeInterval(fromGMT), expectedResult)
+        XCTAssertEqual(iso8601DateOnly?.addingTimeInterval(fromGMT), expectedResult.addingTimeInterval(-fromStartOfDay))
+        XCTAssertEqual(iso8601DateHrMinOnly, expectedResult)
+        XCTAssertEqual(iso8601DateHrMinSOnly, expectedResult)
+        XCTAssertEqual(iso8601DateHrMinSMs, expectedResult)
         XCTAssertNil(invalid)
     }
 
--- a/EZSwiftExtensionsTests/DateTests.swift
+++ b/EZSwiftExtensionsTests/DateTests.swift
@@ -38,28 +38,29 @@ class DateTests: XCTestCase {
 
     func testHTTPDateString() {
         // Given
-        let expectedResult = Date(timeIntervalSince1970: 784_111_777)
+        let expectedResult = Date(timeIntervalSince1970: 1447517111)
         let fromGMT = TimeInterval(NSTimeZone.local.secondsFromGMT())
-        let fromStartOfDay = TimeInterval(8 * 3600 + 49 * 60 + 37) // seconds from start of day
+        let fromStartOfDay = TimeInterval(16 * 3600 + 5 * 60 + 11) // seconds from start of day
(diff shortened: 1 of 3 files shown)


EZSwiftExtensionsTests/DateTests.swift
```diff
@@ -40,17 +40,26 @@ class DateTests: XCTestCase {
         // Given
         let expectedResult = Date(timeIntervalSince1970: 784_111_777)
         let fromGMT = TimeInterval(NSTimeZone.local.secondsFromGMT())
+        let fromStartOfDay = TimeInterval(8 * 3600 + 49 * 60 + 37) // seconds from start of day
 
         // When
         let rfc1123 = Date(httpDateString: "Sun, 06 Nov 1994 08:49:37 GMT")
         let rfc850 = Date(httpDateString: "Sunday, 06-Nov-94 08:49:37 GMT")
         let asctime = Date(httpDateString: "Sun Nov  6 08:49:37 1994")
+        let iso8601DateOnly = Date(httpDateString: "1994-11-06")
```
Khalian: 1994-11-06 is used in multiple places. I think we can abstract that out as a base date. 

I am curious as to what this date corresponds to? I only know of 1970 Jan 1st (which is the beginning of unix time). 

If there is nothing significant here I presume we can use some standard date.

Sources/DateExtensions.swift
```diff
@@ -32,10 +32,26 @@ extension Date {
...
         }
+        if let iso8601DateOnly = Date(fromString: httpDateString, format: "yyyy-MM-dd") {
+            self = iso8601DateOnly
+            return
+        }
+        if let iso8601DateHrMinOnly = Date(fromString: httpDateString, format: "yyyy-MM-dd'T'HH:mmZZZZ") {
+            self = iso8601DateHrMinOnly
+            return
+        }
+        if let iso8601DateHrMinSOnly = Date(fromString: httpDateString, format: "yyyy-MM-dd'T'HH:mm:ssZZZZ") {
```
Khalian: MinS is not descriptive. MinSec is better.

Sources/DateExtensions.swift
```diff
@@ -32,10 +32,26 @@ extension Date {
...
+            self = iso8601DateHrMinSOnly
+            return
+        }
+        if let iso8601DateHrMinSMs = Date(fromString: httpDateString, format: "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZ") {
```
Khalian: *MinSecMillSec or *MinSecMs.

EZSwiftExtensionsTests/DateTests.swift
```diff
@@ -40,17 +40,26 @@ class DateTests: XCTestCase {
...
+        let iso8601DateHrMinSMs = Date(httpDateString: "1994-11-06T08:49:37.000+00:00")
         let invalid = Date(httpDateString: "1994-11-06 08:49:37")
 
         // Test
         XCTAssertEqual(rfc1123, expectedResult)
         XCTAssertEqual(rfc850, expectedResult)
         XCTAssertEqual(asctime?.addingTimeInterval(fromGMT), expectedResult)
+        XCTAssertEqual(iso8601DateOnly?.addingTimeInterval(fromGMT), expectedResult.addingTimeInterval(-fromStartOfDay))
+        XCTAssertEqual(iso8601DateHrMinOnly, expectedResult)
+        XCTAssertEqual(iso8601DateHrMinSOnly, expectedResult)
```
Khalian: This is a little uncomfortable. if iso8601DateHrMinOnly = expectedResult and iso8601DateHrMinSMs = expectedResult. Then iso8601DateHrMinOnly = iso8601DateHrMinSMs. But they should clearly differ by the second and millisecond granuality. 

Can we rewrite these tests to recognize the granularity differences.

Sources/DateExtensions.swift
```diff
@@ -32,10 +32,26 @@ extension Date {
...
-        if let asctime =  Date(fromString: httpDateString, format: "EEE MMM d HH':'mm':'ss yyyy") {
+        if let asctime = Date(fromString: httpDateString, format: "EEE MMM d HH':'mm':'ss yyyy") {
             self = asctime
             return
         }
+        if let iso8601DateOnly = Date(fromString: httpDateString, format: "yyyy-MM-dd") {
+            self = iso8601DateOnly
+            return
+        }
+        if let iso8601DateHrMinOnly = Date(fromString: httpDateString, format: "yyyy-MM-dd'T'HH:mmZZZZ") {
```
Khalian: Question : what does ZZZZ correspond to ?

Khalian:
Moreover, these changes also need changelog entries.

Sources/DateExtensions.swift
Vic-L: My bad it should be a single 'Z'. I'm surprised that it still converts correctly

EZSwiftExtensionsTests/DateTests.swift
Vic-L: I'm not sure. I took it from the previous test examples

Khalian: Then I presume even those tests are bad. 

Regardless, these new tests should contain some sort of distinction between the different extensions you are proposing, based on the granularity.

Vic-L: I made some new commits. I am using the date time of the first commit of this pod as the base date just for this test scenario.

I won't be using the `string` attribute of this test class as the base date for this test scenario because it is being used by other test cases, which are actually failing in my case due to my time zone.

I believe a rewrite of all the tests to improve their robustness is required and maybe we should do it in another PR?

I also intend to follow up with another PR to take in the locale as a parameter for various date/string related functions. Maybe I can do them together in that PR?

Sources/DateExtensions.swift
Vic-L: My bad again..it should be `xxxxx`. This will return the colon in the zone substring.

EZSwiftExtensionsTests/DateTests.swift
```diff
@@ -11,46 +11,58 @@ import XCTest
...
 
         // When
-        let rfc1123 = Date(httpDateString: "Sun, 06 Nov 1994 08:49:37 GMT")
-        let rfc850 = Date(httpDateString: "Sunday, 06-Nov-94 08:49:37 GMT")
-        let asctime = Date(httpDateString: "Sun Nov  6 08:49:37 1994")
-        let invalid = Date(httpDateString: "1994-11-06 08:49:37")
+        let rfc1123 = Date(httpDateString: "Sat, 14 Nov 2015 16:05:11 GMT")
+        let rfc850 = Date(httpDateString: "Saturday, 14-Nov-15 16:05:11 GMT")
+        let asctime = Date(httpDateString: "Sun Nov 14 16:05:11 2015")
+        let iso8601DateOnly = Date(httpDateString: "2015-11-14")
```
Khalian: Again, 2015-11-14 is reused a bunch of times.

EZSwiftExtensionsTests/DateTests.swift
```diff
@@ -11,46 +11,58 @@ import XCTest
...
+    var invalidDateString: String!
+    var date: Date!
+    var timeZoneOffset: TimeInterval!
+    let format = "yyyy-MM-dd'T'HH:mm:ssxxxxx"
 
     override func setUp() {
         super.setUp()
-        string = "01-01-1970 00:00:00"
-        wrongDateString = "13-82-1900 90:65:12"
+        self.dateString = "2015-11-14T16:05:11+00:00" // time of the first commit of EZSwiftExtension :clap::clap:
```
Khalian: :)

EZSwiftExtensionsTests/DateTests.swift
```diff
@@ -229,67 +241,55 @@ class DateTests: XCTestCase {
...
-        let customDate3 = Date(fromString: "11-11-2012 05:45:12", format: self.format)
-        XCTAssertEqual(customDate3?.monthAsString, "November")
+        XCTAssertEqual(self.date.monthAsString, "November")
     }
     
     func testWeekDay() {
-        let customDate = Date(fromString: "09-01-2015 05:45:12", format: self.format)
-        XCTAssertEqual(customDate?.weekday, "Friday")
+        XCTAssertEqual(self.date.weekday, "Sunday")
+        NSTimeZone.default = TimeZone(abbreviation: "UTC")!
```
Khalian: Does this have something to do with the travis server build times being in greenwich or something ?

Vic-L: I am setting a default timmezone because the tests are failing for me because Im on another time zone. So the tests on `day`, `hour` etc variables of the Date extension are failing.

I thought that forcing a default timezone to control the setup of the environment is the way to go.

Please take a look at `testHour()` and `testDay()` test cases. The results differ when the default time zone changed and I am thinking this makes the tests better.

What do you think?

EZSwiftExtensionsTests/DateTests.swift
Vic-L: It is to simulate a json string, and it needs to match the base `dateString`.

Do you mean we should put this string into a variable and and then append the remaining parts of the date string as necessary?

EZSwiftExtensionsTests/DateTests.swift
Khalian: Good call. I think when I wrote those tests I only cared about the build server. This looks better.

EZSwiftExtensionsTests/DateTests.swift
Khalian: Yea, its really a minor thing.

Sources/DateExtensions.swift
```diff
@@ -32,10 +32,26 @@ extension Date {
...
-        if let asctime =  Date(fromString: httpDateString, format: "EEE MMM d HH':'mm':'ss yyyy") {
+        if let asctime = Date(fromString: httpDateString, format: "EEE MMM d HH':'mm':'ss yyyy") {
             self = asctime
             return
         }
+        if let iso8601DateOnly = Date(fromString: httpDateString, format: "yyyy-MM-dd") {
+            self = iso8601DateOnly
+            return
+        }
+        if let iso8601DateHrMinOnly = Date(fromString: httpDateString, format: "yyyy-MM-dd'T'HH:mmxxxxx") {
```
Khalian: What does xxxx stand for?

Vic-L: B the way its 5 x.
It is to produce `+00:00` format for the timezone.
`ZZZZ` produces `+0000` without the `:`.

[Apple docs](https://developer.apple.com/reference/foundation/dateformatter) gave an example that `ZZZZZ` will give `+00:00` with the `:` but that does not seem to be the case.

I am getting this error with `ZZZZZ` for line 36:
`XCTAssertEqual(dateFromString.toString(format: self.format), self.dateString!)`
The error:
`test failure: -[DateTests testDateFromString()] failed: XCTAssertEqual failed: ("2015-11-14T16:05:11Z") is not equal to ("2015-11-14T16:05:11+00:00") - ` on line 36

I found 5x to work the way I needed.

I cant find anything on the 5 x from official Apple docs though :(

PR #427, opened 2017-05-21
feat: array step function

Vic-L:
## Checklist

- [x] New Extension
- [x] New Test

Inspired from ruby's [step](https://ruby-doc.org/core-2.2.0/Range.html#method-i-step)

### DIFF
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -12,12 +12,19 @@ import EZSwiftExtensions
 class ArrayTests: XCTestCase {
     let emptyArray = [Int]()
     var numberArray = [Int]()
+    var multiDimensionalArray = [[Int]]()
 
     override func setUp() {
         super.setUp()
         //[0, 1, 2, 3, 4, 5, 1]
         numberArray = [Int](0...5)
         numberArray.append(1)
+        
+        //[[0, 1], [2, 3], [4, 5], [1]]
+        multiDimensionalArray.append([0,1])
+        multiDimensionalArray.append([2,3])
+        multiDimensionalArray.append([4,5])
+        multiDimensionalArray.append([1])
     }
     
     func testGet_SubArray() {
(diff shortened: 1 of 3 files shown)


Sources/ArrayExtensions.swift
```diff
@@ -196,6 +196,14 @@ extension Array where Element: Equatable {
     public func unique() -> Array {
         return reduce([]) { $0.contains($1) ? $0 : $0 + [$1] }
     }
+    
+    /// EZSE: Break array into arrays of desired step (inspired by ruby's `step`)
+    public func step(_ stepSize: Int) -> [[Element]]? {
```
Khalian: I do not understand this.

Vic-L: Ah its to break an array into smaller arrays of specified size.

Use case: I have a view with 2 columns of check boxes, and I am populating it from an array returned from the backend. The array does not have a fixed size, so I have to load xib into the view.

I have to go row by row and add 2 checkboxes to each row, before moving to the next 1, so I had the need to break it up into arrays of smaller chunk

Khalian: So it the array is k dimensional, and if a programmer performs step(n) on it, does it create n k-1 dimensional arrays in row echelon form? 

If that is indeed the case, we would require some test cases on multi dimensional arrays.

EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -285,6 +285,24 @@ class ArrayTests: XCTestCase {
...
     }
     
+    func testStep() {
+        XCTAssertNil(self.numberArray.step(-1))
+        XCTAssertNil(self.numberArray.step(0))
+
+        XCTAssertEqual(self.numberArray.step(2)![0], [0, 1])
+        XCTAssertEqual(self.numberArray.step(2)![1], [2, 3])
+        XCTAssertEqual(self.numberArray.step(2)![2], [4, 5])
+        XCTAssertEqual(self.numberArray.step(2)![3], [1])
```
Khalian: Why is this value 1?

EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -285,6 +285,24 @@ class ArrayTests: XCTestCase {
...
+        
+        XCTAssertEqual(self.numberArray.step(3)![0], [0, 1, 2])
+        XCTAssertEqual(self.numberArray.step(3)![1], [3, 4, 5])
+        XCTAssertEqual(self.numberArray.step(3)![2], [1])
```
Khalian: And this value as well?

EZSwiftExtensionsTests/ArrayTests.swift
Vic-L: In the `setup()` function, the numberArray == `[0, 1, 2, 3, 4, 5, 1]`

There are 7 digits so the last array will have only 1 entry for steps of 2 or 3.

I think `[1]` will be better than `[1, nil]` or `[1, nil, nil]`

Sources/ArrayExtensions.swift
Vic-L: Right, I didnt consider these cases. Let me review the tests. Thanks!

EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -285,6 +292,54 @@ class ArrayTests: XCTestCase {
...
+        
+        XCTAssertEqual(self.numberArray.step(7)!.count, 1)
+        XCTAssertEqual(self.numberArray.step(7)![0].count, self.numberArray.count) // count in each of the first child element will be <= stepSize
+        XCTAssertEqual(self.numberArray.step(7)![0], self.numberArray)
+        
+        XCTAssertEqual(self.numberArray.step(8)!.count, 1)
+        XCTAssertEqual(self.numberArray.step(8)![0].count, self.numberArray.count) // count in each of the first child element will be <= stepSize
+        XCTAssertEqual(self.numberArray.step(8)![0], self.numberArray)
+        
+        // multidimensional array [[0, 1], [2, 3], [4, 5], [1]]
```
Khalian: It would be better if the multi dimensional array tests live in a separate method. (Not a strong preference though)

EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -285,6 +292,54 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(duplicatedElementsArrayOrderMaintained.unique(), [2, 1, 3])
     }
     
+    func testStep() {
+        XCTAssertNil(self.numberArray.step(-1))
+        XCTAssertNil(self.numberArray.step(0))
+
+        XCTAssertEqual(self.numberArray.step(2)!.count, 4)
```
Khalian: This is hard to read. Can you make some comments which describe the output of the step just prior to the count and array validations.

EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -285,6 +292,54 @@ class ArrayTests: XCTestCase {
...
+        XCTAssertEqual(self.numberArray.step(2)![0].count, 2) // count in each of the first child element will be <= stepSize
```
Khalian: In hindsight the count verifications are superflous, since we are also checking the subarray.

Vic-L:
Hmm im not sure how to implement this for multi dimensional array. Will close this PR

PR #429, opened 2017-05-27
Text field validations

Vic-L:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensionsTests/UITextFieldTests.swift
+++ b/EZSwiftExtensionsTests/UITextFieldTests.swift
@@ -74,8 +74,9 @@ class UITextFieldTests: XCTestCase {
         
         // valid format
         textField.text = "lol@lol.lol"
+        let acceptableSpecialChars = "!#$%&'*+-/=?^_`{|}~".characters
         XCTAssertTrue(textField.validateEmail())
-        for specialChar in "!#$%&'*+-/=?^_`{|}~".characters {
+        for specialChar in acceptableSpecialChars {
             textField.text = "lol\(specialChar)lol@lol.lol"
             XCTAssertTrue(textField.validateEmail())
             textField.text = "\(specialChar)lol@lol.lol"
@@ -84,6 +85,16 @@ class UITextFieldTests: XCTestCase {
             XCTAssertTrue(textField.validateEmail())
             textField.text = "lol\(specialChar)@lol.lol"
             XCTAssertTrue(textField.validateEmail())
+            for anotherSpecialChar in acceptableSpecialChars {
+                textField.text = "lol\(anotherSpecialChar)\(anotherSpecialChar)lol@lol.lol"
+                XCTAssertTrue(textField.validateEmail())
+                textField.text = "\(anotherSpecialChar)lol\(anotherSpecialChar)@lol.lol"
+                XCTAssertTrue(textField.validateEmail())
+                textField.text = "\(anotherSpecialChar)\(anotherSpecialChar)@lol.lol"
+                XCTAssertTrue(textField.validateEmail())
+                textField.text = "lol\(anotherSpecialChar)lol\(specialChar)@lol.lol"
+                XCTAssertTrue(textField.validateEmail())
+            }
         }
         textField.text = "lol.lol@lol.lol"
         XCTAssertTrue(textField.validateEmail())
@@ -144,6 +155,8 @@ class UITextFieldTests: XCTestCase {
         textField.text = "lol123"
         XCTAssertFalse(textField.validateDigits())
         textField.text = "123lol"
+        XCTAssertFalse(textField.validateDigits())
+        textField.text = "1l2o3l"
         XCTAssertFalse(textField.validateDigits())
     }
 }
--- a/EZSwiftExtensionsTests/UITextFieldTests.swift
+++ b/EZSwiftExtensionsTests/UITextFieldTests.swift
@@ -15,6 +15,7 @@ class UITextFieldTests: XCTestCase {
     let y: CGFloat = 35
     let width: CGFloat = 40
     let height: CGFloat = 45
+    let testLength = 5
     
     func testFrame() {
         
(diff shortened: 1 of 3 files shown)


Sources/UITextFieldExtensions.swift
```diff
@@ -47,3 +47,43 @@ extension UITextField {
...
+            return self.text!.characters.count < count
+        case .lessThanOrEqualTo:
+            return self.text!.characters.count <= count
+        }
+    }
+
+    /// EZSE: Validation of email format based on https://stackoverflow.com/questions/201323/using-a-regular-expression-to-validate-an-email-address and https://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-an-email-address
+    // TODO match String.isEmail method
+    func validateEmail() -> Bool {
+        let emailRegEx = "(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\\[(?:(?:(2(5[0-5]|[0-4][0-9])|1[0-9][0-9]|[1-9]?[0-9]))\\.){3}(?:(2(5[0-5]|[0-4][0-9])|1[0-9][0-9]|[1-9]?[0-9])|[a-z0-9-]*[a-z0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])"
```
Khalian: You can abstract out this regex as a member of the extension class. That way you wont have to create it over and over again while calling to validate the email.

Sources/UITextFieldExtensions.swift
```diff
@@ -47,3 +47,43 @@ extension UITextField {
     }
 
 }
+
+
+/// MARK: Validations
+extension UITextField {
+    /// EZSE: Ways to validate by comparison
+    enum textFieldValidationOptions: Int {
+        case equalTo, greaterThan, greaterThanOrEqualTo, lessThan, lessThanOrEqualTo
```
Khalian: Minor : Case elems should be on different lines.

Sources/UITextFieldExtensions.swift
```diff
@@ -47,3 +47,43 @@ extension UITextField {
...
+    }
+    
+    /// EZSE: Validation length of character counts in UITextField
+    func validateLength(ofCount count: Int, option: UITextField.textFieldValidationOptions) -> Bool {
```
Khalian: What is the use case for this extension?

EZSwiftExtensionsTests/UITextFieldTests.swift
```diff
@@ -31,4 +32,118 @@ class UITextFieldTests: XCTestCase {
...
+        XCTAssertFalse(textField.validateDigits())
+        textField.text = "lol"
+        XCTAssertFalse(textField.validateDigits())
+        
+        // alphanumerics
+        textField.text = "lol123"
+        XCTAssertFalse(textField.validateDigits())
+        textField.text = "123lol"
+        XCTAssertFalse(textField.validateDigits())
+    }
```
Khalian: Can you add test case of form "1l2o3l"

EZSwiftExtensionsTests/UITextFieldTests.swift
```diff
@@ -31,4 +32,118 @@ class UITextFieldTests: XCTestCase {
...
+        let textField = UITextField(x: x, y: y, w: width, h: height, fontSize: 38)
+        
+        // empty
+        textField.text = ""
+        XCTAssertFalse(textField.validateEmail())
+        
+        // valid format
+        textField.text = "lol@lol.lol"
+        XCTAssertTrue(textField.validateEmail())
+        for specialChar in "!#$%&'*+-/=?^_`{|}~".characters {
```
Khalian: I am not following what this loop accomplishes.

EZSwiftExtensionsTests/UITextFieldTests.swift
```diff
@@ -31,4 +32,118 @@ class UITextFieldTests: XCTestCase {
...
+        XCTAssertFalse(textField.validateEmail())
+        textField.text = "@lol.lol"
+        XCTAssertFalse(textField.validateEmail())
+        textField.text = "@lol.lol"
+        XCTAssertFalse(textField.validateEmail())
+        textField.text = ".@lol.lol"
+        XCTAssertFalse(textField.validateEmail())
+        textField.text = ".lol@lol.lol"
+        XCTAssertFalse(textField.validateEmail())
+    }
```
Khalian: I really like the thorougness of your test cases.

EZSwiftExtensionsTests/UITextFieldTests.swift
Vic-L: Apparently you can have these special characters in the local part of an email address (the one before the `@`).

I was pretty surprise u can have that so this loop ensures each of them are tested at different scenarios (start of the email, end of the email, middle of the email).

I'm missing the scenario where there can be more than 1 of these special character in the same local part of an email. Ill add it in.

Sources/UITextFieldExtensions.swift
Vic-L: Validating password field of min length
Validate min length of a query before make a search

Sources/UITextFieldExtensions.swift
Vic-L: I see. I will put it under the `String` class as
`static let emailRegex = ...`
Is this ok? Is there a need to test this?

Khalian: No it does not have to be in string class. It can be a member in UITextField class (so just put it as a member in this extension).

Khalian:
LGTM. Can you rebase your changes. I will approve and merge post that.

Sources/UITextFieldExtensions.swift
```diff
@@ -47,3 +51,46 @@ extension UITextField {
     }
 
 }
+
+
+/// MARK: Validations
```
Khalian: We can take this comment out.

Khalian:
Erm, rebase not merge.

PR #431, opened 2017-05-28
Converting all UIKit packages to be buildable for only tvOS or IOS.

Khalian:
This would allow EZSwiftExtensions to be usable in other platforms (MacOS and Linux)

## Checklist

- [ ] New Extension
- [ ] New Test

### DIFF
--- a/EZSwiftExtensions/UIStackViewExtensions.swift
+++ b/EZSwiftExtensions/UIStackViewExtensions.swift
@@ -6,6 +6,8 @@
 //  Copyright © 2017 Goktug Yilmaz. All rights reserved.
 //
 
+#if os(iOS) || os(tvOS)
+
 import UIKit
 
 @available(iOS 9.0, *)
@@ -40,3 +42,5 @@ extension UIStackView {
 	}
 
 }
+
+#endif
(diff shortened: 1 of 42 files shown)


Khalian:
@goktugyil I need both your blessing and oversight for this change. 

The work will be 3 fold. 

1. Add the macros to compile parts of the codebase (in particular anything with UI Kit) only for tvOS and IOS (that is what is in this PR).
2. Add a build target for macOSX (next phase). 
3. Add a build target for linux (next phase).

goktugyil:
I think we need input from everyone for this change. I've gone over the code and it looks good for ios/tvos separation. 

However I don't have much experience in macos and no experience in linux. So I can't provide much help beyond this step. You probably know everything I know about the subject.

I think not doing all the changes in 1 major version, but doing the 3 changes over versions would be a more stable way of doing things. (If we mess up somewhere, its easier to find the problem if incremental steps are taken)

Also, exactly how many people were asking for macos? And I don't remember seeing linux?

Khalian:
@goktugyil In that case I will take out the second commit for adding the travis target for macOS. 

There have been some people. Regardless, this is something we should push for.

PR #432, opened 2017-06-08
feat: change shakeViewForTimes function

Vic-L:
## Checklist

- [ ] New Extension
- [ ] New Test

The previous implementation of `shakeViewForTimes(_:)` causes the view to be misaligned after the animation has ended. This implementation does not.

I'm not sure how to test this and there were no previous tests to refer to 😅


Khalian:
I have little context about UIView. @goktugyil Would you mind taking a look ?

goktugyil:
I've never had the misalignment problem. When does it happen?

PR #433, opened 2017-06-12
Macos support for EZSwiftExtensions. 

Khalian:
## Checklist

- [ ] New Extension
- [ ] New Test


Khalian:
@goktugyil - I want to put this change in before I draft the new release. I would appreciate your review for this one, since this is a massive change.

goktugyil:
Code looks good, but I'm worried about the proj and scheme files. Usually you can't catch issues in those without stumbling into it while using :/

goktugyil:
Lets just not make a version for a while, gives us time to catch stuff.

goktugyil:
Also why not use 
`#if !os(macOS)`
?

Khalian:
@goktugyil I agree. I can guarantee you the following.

1. The build works fine. 
2. The target builds fine locally on my XCode. 
3. The tests run fine locally on my XCode. 

How about I keep these changes for a couple of weeks or so? After that I can merge them. 

Regardng !os(macOS), I also have to make changes for linux and maybe watchOS depending on demand. In linux, UI Kit is not defined, so this change is good.

PR #440, opened 2017-07-15
Added @discardableResult to ez.runThisEvery()

goktugyil:
- [X] Trivial change (doesn't require changelog)

### DIFF
--- a/Sources/EZSwiftFunctions.swift
+++ b/Sources/EZSwiftFunctions.swift
@@ -254,7 +254,7 @@ public struct ez {
     }
 
     /// EZSE: Runs every second, to cancel use: timer.invalidate()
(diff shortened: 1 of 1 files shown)


Sources/EZSwiftFunctions.swift
```diff
@@ -254,7 +254,7 @@ public struct ez {
     }
 
     /// EZSE: Runs every second, to cancel use: timer.invalidate()
-    public static func runThisEvery(seconds: TimeInterval, startAfterSeconds: TimeInterval, handler: @escaping (CFRunLoopTimer?) -> Void) -> Timer {
+    @discardableResult public static func runThisEvery(seconds: TimeInterval, startAfterSeconds: TimeInterval, handler: @escaping (CFRunLoopTimer?) -> Void) -> Timer {
```
Khalian: Why are we putting this annotation ?

goktugyil: Is expecting the result to be used, but most of the time its not needed. Even when its needed, its usually used inside the handler.

PR #445, opened 2017-08-22
Safe array extension

lfarah:
A common crash around swift is to get to subscript an array with a bigger index than the number of array's objects. This extension fixes that by giving an optional value.

This code is from @nkukushkin's [StackOverflow's answer](https://stackoverflow.com/a/30593673/2938877)

## Checklist

- [x] New Extension
- [ ] New Test

### DIFF
--- a/EZSwiftExtensionsTests/ArrayTests.swift
+++ b/EZSwiftExtensionsTests/ArrayTests.swift
@@ -32,6 +32,14 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(emptyArray.get(at: -3...(-1)), [])
         XCTAssertEqual(emptyArray.get(at: -1...1), [])
     }
+    
+    func testSafeIndex() {
+        
+        let optionalNumber: Int? = 3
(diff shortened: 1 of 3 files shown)


EZSwiftExtensionsTests/ArrayTests.swift
```diff
@@ -32,6 +32,14 @@ class ArrayTests: XCTestCase {
         XCTAssertEqual(emptyArray.get(at: -3...(-1)), [])
         XCTAssertEqual(emptyArray.get(at: -1...1), [])
     }
+    
+    func testSafeIndex() {
+        
+        let optionalNumber: Int? = 3
+        XCTAssertEqual(numberArray[safe: 3], optionalNumber)
+        XCTAssertNotEqual(numberArray[safe: 4], 3)
```
Khalian: Is there an implicit collision safety compared to the standard array subscript impl (e.g. numberArray[3] vs numberArray[safe: 3]). If so is the compile time check on the arg name (in this case safe?)

PR #446, opened 2017-08-22
Added + extension for AttributedStrings

lfarah:
## Checklist

- [x] New Extension
- [ ] New Test

### DIFF
--- a/Sources/NSAttributedStringExtensions.swift
+++ b/Sources/NSAttributedStringExtensions.swift
@@ -81,5 +81,4 @@ public func + (left: NSAttributedString, right: NSAttributedString) -> NSAttribu
     return ns
 }
 
-
 #endif
--- a/Sources/NSAttributedStringExtensions.swift
+++ b/Sources/NSAttributedStringExtensions.swift
@@ -73,5 +73,13 @@ public func += (left: inout NSAttributedString, right: NSAttributedString) {
     ns.append(right)
     left = ns
 }
+    
+/// EZSE: Sum of one NSAttributedString with another NSAttributedString
+public func + (left: NSAttributedString, right: NSAttributedString) -> NSAttributedString {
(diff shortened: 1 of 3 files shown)


Sources/NSAttributedStringExtensions.swift
```diff
@@ -73,5 +73,12 @@ public func += (left: inout NSAttributedString, right: NSAttributedString) {
     ns.append(right)
     left = ns
 }
+    
+/// EZSE: Sum of one NSAttributedString with another NSAttributedString
+public func + (left: NSAttributedString, right: NSAttributedString) -> NSAttributedString {
```
Khalian: Does NSAttributedString support any other operators by default. If yes, then what is the operator precedence of those operators wrt to this +?

lfarah: Yes. It [does currently support **+=**](https://github.com/goktugyil/EZSwiftExtensions/blob/adding_attributed_string_extension/Sources/NSAttributedStringExtensions.swift#L71)

Khalian: Its not really about adding something. I like the extension, I am just confirmed about operator precedence. 

Ideally + should be higher precedence than +=.

For instance str1 += str2 + str3 should evaluate to str1 = (str1 + (str2 + str3))). I do not want the assignment to be done prior to the +. 

I wanted to verify that this behavior is consistent for this.

lfarah: I added one more test. What do you think?

Khalian: yea, this verifies my assumption. I will approve as soon as it passes the build.

lfarah:
@Khalian please merge this and I'll fix the Swiftlint error in another PR

PR #447, opened 2017-08-26
Added Enum extensions

furuyan:
## Checklist

- [x] New Extension
- [ ] New Test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Release 1.10]
 
-### New platform added.
+### New platform added. 
 
 1. **MacOSX 10.11**
 
-### Added extensions
+### Added extensions 
 
 1. **Array**
    - `removeAll(_ elements: [Element])` (new overload) for `Equatable` elements in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/416) by *dfrib*
@@ -52,7 +52,7 @@ All notable changes to this project will be documented in this file.
    - `nakedView()` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/424) by *Vic-L*
 
 9. **UIViewControllerExtensions**
-   - `isNavBarHidden` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/428) by *Khalian*
+   - `isNavBarHidden` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/428) by *Khalian* 
 
 10. **Double**
    - `var toCGFloat` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/425) by *Vic-L*
@@ -68,7 +68,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Release 1.9]
 
-### Fixed bugs
+### Fixed bugs 
 
 1. **Array**
    - `foreach` is now removed since its already in the std library by *Khalian*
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -117,7 +122,7 @@ All notable changes to this project will be documented in this file.
    - `set(text _text: String?, duration: TimeInterval)` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/340) by *vilapuigvila*
 
 11. **UIImage**
-   - `var base64: String` in [[PR]](https://github.com/goktugyil/EZSwiftExtensions/pull/363) by *lfarah*## [Version 1.8]
```
Khalian: Please revert this and all prior spaces.

Sources/EnumExtensions.swift
```diff
@@ -0,0 +1,42 @@
...
+            defer { n += 1 }
+
+            let next = withUnsafePointer(to: &n) {
+                UnsafeRawPointer($0).assumingMemoryBound(to: Case.self).pointee
+            }
+            return next.hashValue == n ? next : nil
+        }
+    }
+
+    /// EZSwiftExtensions
```
Khalian: Please fill out descriptive comments.

EZSwiftExtensionsTests/EnumTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+class EnumTests: XCTestCase {
+    enum Fruit: EnumCollection {
+        case apple, orange, banana
+    }
+
+    func testEnumerated() {
+        for (i, fruit) in Fruit.enumrated() {
+            switch i {
+            case 0:
```
Khalian: How is this more useful for than pattern matching on enums in swift ? : https://appventure.me/2015/08/20/swift-pattern-matching-in-detail/#sec-3-5? The indices hide the order without a lookup. The pattern matching syntax allows a proper destruct on the values and is vastly more readable. 

I do not think I like this extension, since there is already a cleaner way to match on enums in the language.

Sources/EnumExtensions.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+extension EnumCollection where Case: Hashable {
+    fileprivate static var iterator: AnyIterator<Case> {
+        var n = 0
+        return AnyIterator {
+            defer { n += 1 }
+
+            let next = withUnsafePointer(to: &n) {
+                UnsafeRawPointer($0).assumingMemoryBound(to: Case.self).pointee
```
Khalian: the three worst words in Programming languages. Unsafe, raw and pointer. :P

What is the expectation on calling an index that is outside of an enum specificiation (e.g. in your example case 3?).

CHANGELOG.md
furuyan: I'm Sorry, fixed.

Sources/EnumExtensions.swift
furuyan: In Swift, `UnsafePointer` means it's just not under the ARC.
And `raw` 's lifetime is limited in  `withUnsafePointer` block.
So I think it doesn't matter.

>What is the expectation on calling an index that is outside of an enum specificiation (e.g. in your example case 3?).

It expects nil.
`let castedNext = next as? Case // castedNext is nil`

Khalian: Whats the garbage collection expectaion on the next pointer here then, if its not covered by ARC?

EZSwiftExtensionsTests/EnumTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+class EnumTests: XCTestCase {
+    enum Fruit: EnumCollection {
+        case apple, orange, banana
+    }
+
+    func testEnumerated() {
+        for (i, fruit) in Fruit.enumerated() {
+            switch (i, fruit) {
+            case (0, .apple),
```
Khalian: I still do not grok why this is any better than pattern matching.

Sources/EnumExtensions.swift
furuyan: According to the `withUnsafePointer`'s document, 
`The pointer argument is valid only for the duration of the function’s execution.`
And besides UnsafePointer is constructed by struct, so we don't need to think about it.

PR #453, opened 2017-09-07
Added public to .base64, so it's available in projects

lfarah:
screen shot 2017-09-07 at 2 51 42 pm
## Checklist

- [ ] New Extension
- [ ] New Test

### DIFF
--- a/Sources/UIImageExtensions.swift
+++ b/Sources/UIImageExtensions.swift
@@ -13,7 +13,7 @@ import UIKit
 extension UIImage {
     
     /// EZSE: Returns base64 string
-    var base64: String {
+    public var base64: String {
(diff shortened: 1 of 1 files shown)


Sources/UIImageExtensions.swift
```diff
@@ -13,7 +13,7 @@ import UIKit
 extension UIImage {
     
     /// EZSE: Returns base64 string
-    var base64: String {
+    public var base64: String {
```
Khalian: Cool. I presume there are unit tests already written for this?

lfarah: It's actually the only test for UIImage: https://github.com/goktugyil/EZSwiftExtensions/blob/master/EZSwiftExtensionsTests/UIImageTests.swift#L16

Issue #463, opened 2017-10-20
UIDeviceExtensions DeviceList


Adrift001:
no iPhone 8, iPhone 8P, iPhone X.

PR #464, opened 2017-10-25
Updated Device Mapping for Newer Iphones

pnishanth:
## Checklist

- [ ] New Extension
- [ ] New Test

#463

### DIFF
--- a/Sources/UIDeviceExtensions.swift
+++ b/Sources/UIDeviceExtensions.swift
@@ -26,6 +26,9 @@ private let DeviceList = [
     /* iPhone 7 */        "iPhone9,1": "iPhone 7", "iPhone9,3": "iPhone 7",
     /* iPhone 7 Plus */   "iPhone9,2": "iPhone 7 Plus", "iPhone9,4": "iPhone 7 Plus",
     /* iPhone SE */       "iPhone8,4": "iPhone SE",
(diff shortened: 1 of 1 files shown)


Sources/UIDeviceExtensions.swift
```diff
@@ -26,6 +26,9 @@ private let DeviceList = [
     /* iPhone 7 */        "iPhone9,1": "iPhone 7", "iPhone9,3": "iPhone 7",
     /* iPhone 7 Plus */   "iPhone9,2": "iPhone 7 Plus", "iPhone9,4": "iPhone 7 Plus",
     /* iPhone SE */       "iPhone8,4": "iPhone SE",
+    /* iPhone 8 */        "iPhone10,1": "iPhone 8", "iPhone10,4": "iPhone 8",
```
Khalian: I have never touched this part of the codebase. Would you mind telling me what the key and value for this mapping semantically represent?

pnishanth: The Key is the hardware string that uniquely identifies every iDevice out there. The mapping is done in order for the user to get "iPhone 8" instead of "iPhone10,1" (Which might be confusing).

Khalian: Cool. I am not going to go out and verify the mappings, so I am approving this as is.

PR #473, opened 2018-03-22
Minor crash resolved

ioskunal:
invalid check mistake resolved

## Checklist

- [ ] New Extension
- [ ] New Test

### DIFF
--- a/Sources/StringExtensions.swift
+++ b/Sources/StringExtensions.swift
@@ -438,7 +438,7 @@ extension String {
     
     /// EZSE: Checks if String contains Emoji
     public func includesEmoji() -> Bool {
-        for i in 0...length {
+        for i in 0..<length {
             let c: unichar = (self as NSString).character(at: i)
             if (0xD800 <= c && c <= 0xDBFF) || (0xDC00 <= c && c <= 0xDFFF) {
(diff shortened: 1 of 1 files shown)


goktugyil:
Is there an associated test with this? Did it run before? Does it run now?

ioskunal:
Thanks for quick reply. I tested this manually. The loop has to be executed from 0 to n-1 where as it runs from 0 to n and then it crashes at `n` index. I had checked this and now it works fine.

goktugyil:
Could you edit the test cases here so it gives an asset while running the code:

`for i in 0...length {`

But doesn't give an assert with:

`for i in 0..<length {`

https://github.com/goktugyil/EZSwiftExtensions/blob/master/EZSwiftExtensionsTests/StringTests.swift

Sources/StringExtensions.swift
```diff
@@ -438,7 +438,7 @@ extension String {
     
     /// EZSE: Checks if String contains Emoji
     public func includesEmoji() -> Bool {
-        for i in 0...length {
+        for i in 0..<length {
```
Khalian: Could you write a unit test for this pls ?

PR #477, opened 2018-05-15
Swift4finalmergecandidate minor fix

isimple4:
## Checklist

- [ ] New Extension
- [x] New Test

This PR mainly reverts some improper changes back to original, and fixes couple of warnings based on #476 .

Some changes explained as below.

- `SWIFT_SWIFT3_OBJC_INFERENCE` is set to `Default` as suggested [here](https://stackoverflow.com/questions/44379348/the-use-of-swift-3-objc-inference-in-swift-4-mode-is-deprecated)
- The  `Hashable` and `Equatable` thing seems like an [issue](https://bugs.swift.org/browse/SR-5072) with XCode 9.2 compiler, I've revert them back and update to 9.3 for CI. 
- `#if targetEnvironment(simulator)` is suggested to replace `#if (arch(i386) || arch(x86_64)) && os(iOS)` by compiler. [Ref](https://github.com/apple/swift-evolution/blob/master/proposals/0190-target-environment-platform-condition.md)
- Block based `NotificationCenter.​addObserver​(forName:​object:​queue:​using:)` is warned by lint requiring an object to store the returned result, so that it can be removed later. I simply store it to an empty object, but [here](https://oleb.net/blog/2018/01/notificationcenter-removeobserver/) is a better solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-osx_image: xcode9.2
+osx_image: xcode9.3
 language: objective-c
 rvm: 2.2.2
 cache: cocoapods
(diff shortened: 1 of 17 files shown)


isimple4:
@Khalian @goktugyil @lfarah

isimple4:
@EZSwiftExtensionsBot plz test

PR #485, opened 2018-08-01
Implement Hashable hashString

rugheid:
## Checklist
- [x] New Extension
- [x] New Test

### DIFF
--- a/EZSwiftExtensions.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
+++ b/EZSwiftExtensions.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
+</dict>
+</plist>
--- a/EZSwiftExtensionsTests/HashableTests.swift
+++ b/EZSwiftExtensionsTests/HashableTests.swift
@@ -1,0 +1,19 @@
+//
+//  HashableTests.swift
+//  EZSwiftExtensions
+//
+//  Created by Rugen Heidbuchel on 01/08/2018.
+//  Copyright © 2018 Goktug Yilmaz. All rights reserved.
+//
+
+import XCTest
+@testable import EZSwiftExtensions
+
+class HashableTests: XCTestCase {
+    
+    func testHashString() {
+        let s = "test"
+        XCTAssertEqual(s.hashString, String(s.hashValue))
+    }
+    
+}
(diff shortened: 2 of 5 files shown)


EZSwiftExtensions.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Khalian: Why do we need this file?

EZSwiftExtensionsTests/HashableTests.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Created by Rugen Heidbuchel on 01/08/2018.
+//  Copyright © 2018 Goktug Yilmaz. All rights reserved.
+//
+
+import XCTest
+@testable import EZSwiftExtensions
+
+class HashableTests: XCTestCase {
+    
+    func testHashString() {
```
Khalian: Lets also add a negative test case here (i.e. a situation in which XCTAssertNotEqual(s.hashString, String(s.hashValue)). 

But I cannot think of any such strings of the top of my head. Let me know if thats the case.

Khalian: I see. I still have to do the work to move it to swift 4.1 (its right now in swift 4.0 with XCode 9.2 as the build). 

I presume this file is a no op for prior XCode builds?

EZSwiftExtensions.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
Khalian: I see. I still have to do the work to move it to swift 4.1 (its right now in swift 4.0 with XCode 9.2 as the build).

I presume this file is a no op for prior XCode builds?

Also while this is recommended in the release notes, I am still very skeptical of generated files. I presume an update to the unit tests would actually be captured here.

Sources/HashableExtensions.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Rugen Heidbuchel on 01/08/2018.
+//  Copyright © 2018 Goktug Yilmaz. All rights reserved.
+//
+
+import Foundation
+
+extension Hashable {
+    
+    /// EZSE: A string representation of the hash value.
+    var hashString: String {
```
Khalian: This is a minor thing, but can this go into StringExtensions? It makes sense to have a Hashable extension class if there are disparate types you want to deem hashable.

Khalian: Lets also add a CHANGELOG entry for this.

rugheid: I think it's better to keep it as a `Hashable` extension. If people create their own `Hashable` classes, this will also work. It will also work for other Apple classes, like `NSSet` for example. I will add a CHANGELOG entry already.

rugheid: I added the CHANGELOG entry.

rugheid:
I think this PR is ready for merge now. 😀

PR #486, opened 2018-08-24
Overloaded -timePassed method with language agnostic implementation

110100110101:
The overload of the method -timePassed returns an enum of TimePassed. Instances of TimePassed can be `.year(int)`, `.month(int)`, `.day(int)`, `.hour(int)`, `.minute(int)`, `.second(int)`, and `now`.

Original implementation of the -timePassed method returns a string eg. `5 seconds ago`, `13 years ago`. It is really convenient for the consumer of this method, if the target build is on english, but it is really cumbersome if the target is non-english builds. 

## Checklist

- [ ] New Extension
- [x] New Test

### DIFF
--- a/Sources/DateExtensions.swift
+++ b/Sources/DateExtensions.swift
@@ -9,7 +9,7 @@
 import Foundation
 
 extension Date {
-
+    
     public static let minutesInAWeek = 24 * 60 * 7
 
     /// EZSE: Initializes Date from string and format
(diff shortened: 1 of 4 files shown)


Sources/DateExtensions.swift
```diff
@@ -136,6 +136,30 @@ extension Date {
...
     }
     
+    /// EZSE: Easy creation of time passed String. Can be Years, Months, days, hours, minutes or seconds. Useful for localization
+    public func timePassed() -> TimePassed {
+        
+        let date = Date()
+        let calendar = Calendar.autoupdatingCurrent
+        let components = (calendar as NSCalendar).components([.year, .month, .day, .hour, .minute, .second], from: self, to: date, options: [])
+        
+        if components.year! >= 1 {
```
Khalian: What if you have to represent something like 1 year and 6 months. IMO it should be both TimePassed.year(1) and TimePassed.month(6). Here it would just round up to 1 year.

110100110101: You're right, it would just round up to the largest time unit, because we retained the original behaviour of the first `-timePassed` extension method. The original method would return `"1 year ago"`, instead of `"1 year and 6 months ago"`. 

On developers' (consumers') perspective, you're still right on what the `-timePassed` method should return.

PR #499, opened 2019-04-18
Surport Swift 5

ceeyang:
## Checklist

- [x] New Extension
- [x] New Test

### DIFF
--- a/Sources/DateExtensions.swift
+++ b/Sources/DateExtensions.swift
@@ -137,29 +137,29 @@ extension Date {
     }
     
     /// EZSE: Easy creation of time passed String. Can be Years, Months, days, hours, minutes or seconds. Useful for localization
-    public func timePassed() -> TimePassed {
-        
-        let date = Date()
-        let calendar = Calendar.autoupdatingCurrent
-        let components = (calendar as NSCalendar).components([.year, .month, .day, .hour, .minute, .second], from: self, to: date, options: [])
-        
-        if components.year! >= 1 {
-            return TimePassed.year(components.year!)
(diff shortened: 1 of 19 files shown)


Sources/DateExtensions.swift
```diff
@@ -137,29 +137,29 @@ extension Date {
...
-        } else if components.minute! >= 1 {
-            return TimePassed.minute(components.minute!)
-        } else if components.second! >= 1 {
-            return TimePassed.second(components.second!)
-        } else {
-            return TimePassed.now
-        }
-    }
-    
+//    public func timePassed() -> TimePassed {
```
Khalian: Why are we commenting this out?

Sources/IntExtensions.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import AVFoundation
```
Khalian: Why do we need this ?

Sources/NSObjectExtensions.swift
```diff
@@ -6,6 +6,8 @@
 //  Copyright (c) 2015 Goktug Yilmaz. All rights reserved.
 //
 
+import Foundation
```
Khalian: Do we need this ?

Khalian:
I have created a remote branch ceeyang-Swift5.0, I will keep pushing in changes to fix builds and stuff, but I would need your help since I am out of date on all the XCode versions and MacOSX versions for latest swift stuff.

ceeyang:
I will make a new PL

